PR #5, opened 2017-02-11
improvements

SimonDanisch:
... on the way to first working prototype


vchuravy:
We will need to make sure that `CUDArt` works on v0.6 which it currently doesn't due to GC issues.

SimonDanisch:
Is there any issue tracking this?

PR #7, opened 2017-03-25
Get ready for prerelease

SimonDanisch:


SimonDanisch:
@timholy, do you think this will fit nicely into your accelerator based infrastructure?
Would be awesome to implement more things like filtering with e.g. CLFFT and a generic filter kernel which can get compiled for all the different operations ;)

SimonDanisch:
@izaid this might be interesting for you... GPUArrays is now at the point that I was talking about some time ago! :)

PR #9, opened 2017-04-02
Remove depwarns and such!

SimonDanisch:


src/backends/julia/julia.jl
```diff
@@ -65,7 +65,7 @@ nthreads{T, N}(a::JLArray{T, N}) = context(a).nthreads
 
 Base.@propagate_inbounds Base.getindex{T, N}(A::JLArray{T, N}, i::Integer) = A.buffer[i]
 Base.@propagate_inbounds Base.setindex!{T, N}(A::JLArray{T, N}, val, i::Integer) = (A.buffer[i] = val)
-Base.linearindexing{T, N}(::Type{JLArray{T, N}}) = Compat.IndexLinear()
+@compat Base.IndexStyle{T, N}(::Type{<:JLArray{T, N}}) = IndexLinear()
```
musm: This gives
```julia
WARNING: static parameter T does not occur in signature for linearindexing at C:\Users\appveyor\.julia\v0.5\GPUArrays\src\backends\julia\julia.jl:68.
The method will not be callable.
WARNING: static parameter N does not occur in signature for linearindexing at C:\Users\appveyor\.julia\v0.5\GPUArrays\src\backends\julia\julia.jl:68.
The method will not be callable.
```
on master

Issue #38, opened 2017-07-19
Initializing Second Device


jekyllstein:
Initializing a backend such as opencl it selects the first device by default as seen in lines 29-38 of opencl.jl
[https://github.com/JuliaGPU/GPUArrays.jl/blob/master/src/backends/opencl/opencl.jl](url)
```julia 
dev = if isempty(devlist)
                devlist = cl.devices(:cpu)
                if isempty(devlist)
                    error("no device found to be supporting opencl")
                else
                    first(devlist)
                end
            else
                first(devlist)
            end
```

Can we add functionality to select from multiple supported devices when initializing.  For example, most laptops with discrete graphics also have integrated graphics as an option.  For these computers the first device is the integrated chip which is normally not what you want to use doing GPU acceleration.

NHDaly:
Not sure how the maintainers feel about this issue, but it affects me as well. I've started a small PR for the OpenCL context at #39.

PR #39, opened 2017-07-22
Allows specifying device when initializing OpenCL

NHDaly:
Adds the ability to choose (and switch) device when initializing the OpenCL context, per #38.

For an example of this in use, see
https://gist.github.com/NHDaly/bf18263fdee4d78bae5c4e1de8ad7208

### DIFF
--- a/src/backends/opencl/opencl.jl
+++ b/src/backends/opencl/opencl.jl
@@ -23,7 +23,7 @@ immutable CLContext <: Context
     device::cl.Device
     context::cl.Context
     queue::cl.CmdQueue
-    function CLContext(device_type = nothing)
+    function CLContext(device_type = nothing; device_idx = 1)
         device = if device_type == nothing
             devlist = cl.devices(:gpu)
             dev = if isempty(devlist)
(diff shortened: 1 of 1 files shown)


NHDaly:
I'm definitely not sure the API I introduced is the best way to do it, but thought I'd get this started, since I was in the same boat as #38: it always selects my built-in graphics card, instead of the (more powerful) standalone one.

SimonDanisch:
Thanks for taking a stab at it... I put the issue last on my list for #37, because it will be simple to implement, but one still needs to put some thought in the API.

I guess what we'll need is more fine grained control!
Like selecting all opencl cpu devices. 
Probably the best way to do this would be to have a device iterator and implement a diverse set of filter operation.
So one could do something like this:
```Julia
contexts = allcontexts(z-> isa(z, OpenCL) && isa(z, GPU))
for context in contexts
    make_current(context)
    ....# schedule calculations on device
end
end
```

NHDaly:
Cool!

I'm a bit confused by your example. What would it mean to make multiple contexts "current"?

PR #75, opened 2017-09-29
fix scale! + some other fixes and tests

SimonDanisch:


SimonDanisch:
adds test for this bug: https://github.com/JuliaGPU/CLBLAS.jl/pull/40

src/broadcast.jl
```diff
@@ -104,10 +104,15 @@ function Base.foreach(func, over::GPUArray, Bs...)
 end
 
 
-arg_length(x::Array) = error("No array in GPU Broadcast supported. Try converting it to a gpu array")
+function arg_length(x::AbstractArray)
+    if !isbits(x)
```
vchuravy: Are union array `pointerfree` or `isbits`. Probably similar to your recent PR to OpenCL.jl

SimonDanisch: good call! to used to using isbits for those things!

PR #80, opened 2017-10-17
add JLArray to the README

ChrisRackauckas:


ChrisRackauckas:
Also exports `JLArray` to make it act more first-class.

SimonDanisch:
Ah sorry, forgot about this! Wanted to add, that JLArray is not really multi threaded ;)
It's really just for debugging, and because threads are yielding worse errors, I wanted to go without them.
It would actually be easier to implement JLArrays with threads :D So maybe that's something we could do at some point... ;)

PR #105, opened 2018-02-21
fix #104

### LINKED ISSUE #104: Broadcast with wrapper types
```
julia> xs = cu([1,2,3])
3-element CuArray{Float32,1}:
 1.0
 2.0
 3.0

julia> xs .* xs'
ERROR: Arrays type contains pointer - this doesn't work on the GPU (So arrays that don't contain pointers to memory, e.g. UnitRanges, would work).
            Try converting it to a gpu array. Found: RowVector{Float32,CuArray{Float32,1}}
```

This is with CuArray's own broadcast kernel disabled.

(note that it's not enough to fix this for `RowVector`; we need something that new wrapper types can hook into).

SimonDanisch:


MikeInnes:
Bump, would be great to have this tagged

PR #129, opened 2018-07-18
Fix broadcast

vchuravy:
CLArray/CUArray will have to define the equivalent of:

```
function CUDAnative.cudaconvert(bc::Broadcast.Broadcasted{Style}) where Style
    Broadcast.Broadcasted{Style}(bc.f, map(CUDAnative.cudaconvert, bc.args), bc.axes)
end

function CUDAnative.cudaconvert(ex::Broadcast.Extruded)
    Broadcast.Extruded(CUDAnative.cudaconvert(ex.x), ex.keeps, ex.defaults)
end
```

but this simplifies broadcast a bunch. Test failures are down to a handful now and a whole bunch of deprecation warnings to go.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,6 +1,6 @@
 environment:
   matrix:
-  - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x64/0.7/julia-0.7.0-alpha-win64.exe"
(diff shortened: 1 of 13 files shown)


vchuravy:
This now passes all tests at least locally, still a whole bunch, of deprecation warnings to go through, 
but I would rather have this merged now so that we can get CuArrays working.

PR #130, opened 2018-07-24
Switch to Int as Integer type

vchuravy:
This fairly straight-forward search-and-replace transformation. 

x-ref: https://github.com/JuliaGPU/CuArrays.jl/pull/99#issuecomment-407295029


SimonDanisch:
Ok let me try this on 0.7! We can likely remove a lot of those `Int.(...)` `Int(...)` conversions, since they were meant to do the Int -> UInt32 conversion ;)

PR #137, opened 2018-08-14
allow tests using Julia v1.0 (Travis)

ranocha:
There are tests using Julia v1.0 on Appveyor but not on Travis. Since these are passing and there have been a lot of commits, it might make sense to release a new version of GPUArrays.jl?


SimonDanisch:
Yeah we should soon make a release!

PR #143, opened 2018-08-17
Clean up and extend coverage for matmul tests

andreasnoack:
Now all four standard element types are exercised as well of all nine combinations of none, transpose, and adjoint. Should we add the 16-bit variants at some point?


vchuravy:
You mean Float16? Those don't work on GPUs (long standing crusade of mine)

PR #148, opened 2018-08-24
Support `Transpose` and `Adjoint` in broadcast better

vchuravy:
@maleadt as discussed earlier today, I didn't quite finish it so the CuArray changes are at https://github.com/JuliaGPU/CuArrays.jl/tree/vc/things

@SimonDanisch thoughts on the backend addition? The issue is that with wrapper types like `Adjoint`/`Transpose`/`SubArray` there is no easy dispatch target anymore so something trait-like might work quite well.

@andreasnoack would it make sense to introduce `Adjoint{T, A} <: Wrapper{T, A<:AbstractArray{T}} <: AbstractArray{T}` into the base hierarchy? Right now we need to enumerate and special treat all the wrapper types.

### DIFF
--- a/src/mapreduce.jl
+++ b/src/mapreduce.jl
@@ -5,7 +5,10 @@
 
 Base.any(A::GPUArray{Bool}) = mapreduce(identity, |, A; init = false)
 Base.all(A::GPUArray{Bool}) = mapreduce(identity, &, A; init = true)
-Base.count(pred, A::GPUArray) = Int(mapreduce(pred, +, A; init = 0))
+
+Base.any(f::Function, A::GPUArray) = mapreduce(f, |, A; init = false)
+Base.all(f::Function, A::GPUArray) = mapreduce(f, &, A; init = true)
+Base.count(pred::Function, A::GPUArray) = Int(mapreduce(pred, +, A; init = 0))
 
 Base.:(==)(A::GPUArray, B::GPUArray) = Bool(mapreduce(==, &, A, B; init = true))
 
--- a/src/abstract_gpu_interface.jl
+++ b/src/abstract_gpu_interface.jl
@@ -106,8 +106,8 @@ end
 #     CUDAnative.__syncthreads()
 # end
 
-
-
+abstract type GPUBackend end
+backend(::Type{T}) where T = error("Can't choose GPU backend for $T")
 
 """
     gpu_call(kernel::Function, A::GPUArray, args::Tuple, configuration = length(A))
@@ -124,7 +124,7 @@ Optionally, a launch configuration can be supplied in the following way:
     2) Pass a tuple of integer tuples to define blocks and threads per blocks!
 
 """
-function gpu_call(kernel, A::GPUArray, args::Tuple, configuration = length(A))
+function gpu_call(kernel, A::AbstractArray, args::Tuple, configuration = length(A))
     ITuple = NTuple{N, Integer} where N
     # If is a single integer, we assume it to be the global size / total number of threads one wants to launch
     thread_blocks = if isa(configuration, Integer)
@@ -148,8 +148,8 @@ function gpu_call(kernel, A::GPUArray, args::Tuple, configuration = length(A))
                 `linear_index` will be inbetween 1:prod((blocks..., threads...))
         """)
     end
-    _gpu_call(kernel, A, args, thread_blocks)
+    _gpu_call(backend(typeof(A)), kernel, A, args, thread_blocks)
 end
 
 # Internal GPU call function, that needs to be overloaded by the backends.
-_gpu_call(f, A, args, thread_blocks) = error("Not implemented")
+_gpu_call(::Any, f, A, args, thread_blocks) = error("Not implemented")
--- a/src/abstractarray.jl
+++ b/src/abstractarray.jl
@@ -52,22 +52,24 @@ end
 
 ## showing
 
-for (atype, op) in
-    [(:(GPUArray), :(Array)),
-     (:(LinearAlgebra.Adjoint{<:Any,<:GPUArray}), :(x->LinearAlgebra.adjoint(Array(parent(x))))),
-     (:(LinearAlgebra.Transpose{<:Any,<:GPUArray}), :(x->LinearAlgebra.transpose(Array(parent(x)))))]
+for (AT, f) in
+    (GPUArray                                  => Array,
+     LinearAlgebra.Adjoint{<:Any,<:GPUArray}   => x->LinearAlgebra.adjoint(Array(parent(x))),
+     LinearAlgebra.Transpose{<:Any,<:GPUArray} => x->LinearAlgebra.transpose(Array(parent(x))),
+     SubArray{<:Any,<:Any,<:GPUArray}          => x->SubArray(Array(parent(x)), parentindices(x))
+    )
   @eval begin
     # for display
-    Base.print_array(io::IO, X::($atype)) =
-        Base.print_array(io,($op)(X))
+    Base.print_array(io::IO, X::$AT) =
+        Base.print_array(io,$f(X))
 
     # for show
-    Base._show_nonempty(io::IO, X::($atype), prefix::String) = 
-        Base._show_nonempty(io,($op)(X),prefix)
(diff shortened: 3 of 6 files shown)


ssz66666:
I wonder if this approach can be adapted to allow arbitrarily nested wrapper types. Currently it doesn't:

```julia
julia> using CuArrays, Adapt

julia> CuArrays.allowscalar(false);

julia> xs = Adapt.adapt(CuArray{Complex{Float32}},[1.0+0.0im]);

julia> ys = transpose(xs)
1×1 LinearAlgebra.Transpose{Complex{Float64},CuArray{Complex{Float64},1}}:
 1.0 + 0.0im

julia> ys .= 1
[ Info: getfield(GPUArrays, Symbol("##19#20"))(), (LinearAlgebra.Transpose{Complex{Float64},CuArray{Complex{Float64},1}}, Base.Broadcast.Broadcasted{Nothing,Tuple{Base.OneTo{Int64},Base.OneTo{Int64}},typeof(identity),Tuple{Int64}})
1×1 LinearAlgebra.Transpose{Complex{Float64},CuArray{Complex{Float64},1}}:
 1.0 + 0.0im


julia> zs = adjoint(ys)
1×1 LinearAlgebra.Adjoint{Complex{Float64},LinearAlgebra.Transpose{Complex{Float64},CuArray{Complex{Float64},1}}}:
Error showing value of type LinearAlgebra.Adjoint{Complex{Float64},LinearAlgebra.Transpose{Complex{Float64},CuArray{Complex{Float64},1}}}:
ERROR: scalar getindex is disabled

julia> zs .= 1
ERROR: scalar setindex! is disabled
```

vchuravy:
I am not sure we can handle the recursive case without injecting an abstraction into the type-hierarchy. But everything gets fuzzy once you start thinking about `transpose(DArray(CuArray))`.

src/mapreduce.jl
```diff
@@ -5,7 +5,10 @@
 
 Base.any(A::GPUArray{Bool}) = mapreduce(identity, |, false, A)
 Base.all(A::GPUArray{Bool}) = mapreduce(identity, &, true, A)
-Base.count(pred, A::GPUArray) = Int(mapreduce(pred, +, 0, A))
+
+Base.any(f::Function, A::GPUArray) = mapreduce(f, |, false, A)
```
ssz66666: Just noticed that `Base.any` and `Base.all` have defined slightly different semantics from the mapreduce-based implementation here when I was trying to incorporate the changes into #145 :
1. They are short-circuiting evaluations in `Base`. This changes semantics when the predicate has side-effects. Maybe we can just make it very clear in documentation that the GPU implementation is not short-circuiting?
2. `Base` implementations allow `missing` values, and deal with them by following [three-valued logic](https://en.wikipedia.org/wiki/Three-valued_logic). Should we follow the `Base` behaviour here? It might end up with something like:

```julia
Base.any(pred, A::GPUArray) = let t =
    begin
        mapreduce(x -> let v = pred(x)
            ismissing(v) ? (false, true) : (v && true, false) end, # enforce v is `Bool` or `missing`
        ((v1,m1),(v2,m2)) -> (v1 || v2, m1 || m2),
        A; init = (false, false))
    end
    t[1] ? true : (t[2] ? missing : false)
end
```

ssz66666:
I was thinking about marking wrapped arrays as GPU compatible by using another wrapper struct, so that we can get a dispatch target. It will be necessary to re-implement all Base. methods that directly access its content, like what is done for `GPUArray` already. Then we might hijack methods that create wrappers (like LinearAlgebra.Transpose) over GPUArray/ wrapped GPU compatible array, to unwrap the marker, apply the wrapper, and then remark it as GPU compatible.

I think this way we might be able to handle nested wrapper since the marker is always at the outside, and not hidden by wrapping.

We will need to special case for all these wrappers we want to support vectorisation, but I think that is okay, because only very few wrappers, like `SubArray` or `Transpose` are really needed and actually easily GPU vectorisable without scalar indexing. 

I've got a minimal example to demonstrate the idea. I'll really appreciate your opinion as I think it would be really useful to get some simple wrappers like `SubArray`, `LinearAlgebra.Transpose` or `SkipMissing` to work with GPUArrays.

```julia
abstract type MyInterface{T} end # like AbstractArray

abstract type MyGPUStruct{T} <: MyInterface{T} end # like GPUArray

struct MyStruct{T} <: MyGPUStruct{T} # like CuArray 
    name::String
    x::T
end

struct MyWrapper{T,AT <: MyInterface{T}} <: MyInterface{T} # like LinearAlgebra.Transpose
    parent::AT
end

struct MyWrapper2{T,AT <: MyInterface{T}} <: MyInterface{T} # any Wrapper that we want it to fall back to base
    parent2::AT
end

# methods in Base to create wrappers, like view, transpose and adjoint
wrap1(x::MyInterface) = MyWrapper(x)
wrap2(x::MyInterface) = MyWrapper2(x)

# all methods in `Base` to be overridden,
# like what is already done with GPUArray itself
function get_name(::MyInterface) 
    error("not implemented")
end

#think of get_name as broadcast, fill!, copyto! or show, etc
# we want to override it for OnGPU later

get_name(o::MyStruct) = begin println("MyStruct"); o.name end
get_name(o::MyWrapper) = begin println("Wrapper"); get_name(o.parent) end
get_name(o::MyWrapper2) = begin println("Wrapper2"); get_name(o.parent2) end

#methods that simply fall back to base
get_x(o::MyStruct) = o.x
get_x(o::MyWrapper) = get_x(o.parent)
get_x(o::MyWrapper) =  get_x(o.parent2)

# our marker wrapper that marks an MyInterface like thing
# to be GPU compatible. It only propagates to the next layer of
# wrapper if only the next layer of wrapping 
# is also GPU compatible 
# type param BE remembers the backend type to be able to
# dispatch to specific GPU kernels
# type param AT reveals the type of underlying array
# so that you can also dispatch on that
struct OnGPU{T,BE<:MyGPUStruct{T},AT<:MyInterface{T}} <: MyInterface{T}
    content::AT
end

# need to override methods that create wrappers we care about,
# e.g. Base.view, transpose, etc.
wrap1(x::BE) where {T,BE <: MyGPUStruct{T}} = begin
    y = MyWrapper(x)
    OnGPU{T,BE,typeof(y)}(y)
end
wrap1(x::OnGPU{T,BE,AT}) where {T,BE,AT} = begin
    y = MyWrapper(x.content)
    OnGPU{T,BE,typeof(y)}(y)
end

# capture vector operation methods like fill!
# for broadcast to work we need to fix 
# BroadcastStyle(::Type{OnGPU{T,<:MyGPUStruct,<:MyInterface}})
get_name(x::OnGPU) = get_name_gpu(x.content)

# methods that fall back to base
get_x(x::OnGPU) = get_x(x.content)

# GPU methods which are generic enough to be handled in
# GPUArrays rather than in backend
# in this case, simply unwrapping MyWrapper and pass it onto
# get_name_gpu(::MyGPUStruct)
get_name_gpu(x::MyWrapper) = get_name_gpu(x.parent)

# default behaviour should be falling back to base
get_name_gpu(x::MyGPUStruct) = error("fall back to base")

# hopefully implemented in GPU backend, don't know how difficult
# it would be to handle recursively nested wrappers that wrap
# a GPUArray(CuArray).
get_name_gpu(x::MyStruct) = get_name(x)

x = MyStruct("x",1) # uses gpu method
y = wrap1(x) # dispatch to gpu method
z = wrap1(y) # dispatch to gpu method
w = wrap2(z) # fall back to base for non-compatible wrappers 
v = wrap1(w) # fall back to base again

get_name(x)
get_name(y)
get_name(z)
get_name(w)
get_name(v)
```

ssz66666:
Agreed with@maleadt,  solving the whole wrapper support design problem seems out of scope of this PR.
 
Also, my proposal does break functions expecting a wrapper directly. An obvious use case is `LinearAlgebra.transpose(::Transpose)`. Guess we need to find a way to expose this information to compiler without messing too much with the type hierarchy. Or perhaps just sticking to a simpler design and supporting the simplest yet common cases, at least for now.

vchuravy:
When we tag we should bump the major version due to the change in API

PR #206, opened 2019-09-13
fix non-deterministic mapreduce compilation error and inbounds

ChrisRackauckas:
DiffEqGPU was getting non-deterministic `simple_broadcast_index` compilation errors which fixing the inputs and inbounds seems to fix.


vchuravy:
In particular we were seeing a call to `simple_broadcast_index(A, i, i)`

ChrisRackauckas:
Indeed, and how the matrix broadcasts in DiffEqGPU ever worked before when the splatting clearly would call a two index method is quite bizarre 🤷‍♂

PR #428, opened 2022-09-23
Add backend function for Adapt.WrappedArray

vchuravy:


vchuravy:
@maleadt Not sure what the best way for testing this is, but since GPUArraysCore already has a dependency on Adapt this seemed like a sensible place to define it.

PR #437, opened 2022-11-11
Revamp diagonal `cholesky` method

simsurace:
This tries to address no. 1 of #434 and also make some overall improvements.
It assumes `findfirst` is implemented for JLArrays without triggering scalar indexing. In practice, this means it won't pass tests before #436 has been merged.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -2,19 +2,18 @@ steps:
   - label: "CUDA.jl"
     plugins:
       - JuliaCI/julia#v1:
-          version: 1.7
+          version: 1.8
       - JuliaCI/julia-coverage#v1:
           codecov: true
     command: |
-      julia -e 'using Pkg;
+      julia -e 'using Pkg
 
-                println("--- :julia: Instantiating project");
-                Pkg.develop(PackageSpec(path=pwd()));
-                Pkg.add(PackageSpec(name="CUDA", rev="master"));
-                Pkg.instantiate();
+                println("--- :julia: Instantiating project")
+                Pkg.develop(; path=pwd())
+                Pkg.develop(; name="CUDA")
 
-                println("+++ :julia: Running tests");
-                Pkg.test("CUDA"; coverage=true);'
+                println("+++ :julia: Running tests")
+                Pkg.test("CUDA"; coverage=true)'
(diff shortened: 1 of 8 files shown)


maleadt:
> In practice, this means it won't pass tests before #436 has been merged.

You can base PRs on top of other branches that are not master. I've rebased your PR, but there's some other CI failures that look related.

simsurace:
The `nothing`s seem to happen with JLArrays. Unfortunately I don't think I have any means to debug this locally.

maleadt:
> Unfortunately I don't think I have any means to debug this locally.

JLArrays is the test CPU back-end; it's available as the registered JLArrays.jl package (but also available in `lib/JLArrays`) and can also be constructed with the `jl` function (much like `cu`). You don't need any special hardware for it.

simsurace:
It turned out to be a rather silly translation mistake. Let's see whether this does the job.

simsurace:
Ok, this seems to do the trick.

simsurace:
@maleadt, are there any blockers?

maleadt:
Yeah, oneAPI CI fails, see https://github.com/JuliaGPU/GPUArrays.jl/pull/436.

(also, for future reference, please rebase instead of merging; the timeline here now includes unrelated commits)

maleadt:
You introduced even more commits :-) I'll rebase it when I get the time to fix #436.

simsurace:
I'm sorry, I just used the GitHub merge editor to resolve the conflicts that were reported. I don't see any additional commits in this timeline. I guess I'm unfamiliar with the way in which you manage the different branches here.
I'm used to just setting the base branch to master and merging master into the branch before making the merge, which is compatible with the GitHub interface, but presumably you are doing it differently here. Again, apologies for making your life harder than it needs to be.

maleadt:
> I don't see any additional commits in this timeline.

There's a whole lot of unrelated commits in here, no?

image

Granted, they don't cause actual diff in the review tab, so aren't much of an issue (especially if squash merging), but they're a bit messy. Nothing to worry about, again, I'll just rebase when I fix the underlying issue 🙂

maleadt:
Apparently you confused GitHub too, because merging #436 closed this PR... I'll open a new one

PR #504, opened 2023-12-07
Add a GPU version of copytrito!

amontoison:
@michel2323
I added a function in Julia (https://github.com/JuliaLang/julia/pull/51909/) to easily copy the triangular part of a matrix.
This PR adds a fallback for GPU matrices.

### DIFF
--- a/src/host/linalg.jl
+++ b/src/host/linalg.jl
@@ -129,7 +129,7 @@ end
 ## copy a triangular part of a matrix to another matrix
 
 if isdefined(LinearAlgebra, :copytrito!)
-    function copytrito!(B::AbstractGPUMatrix, A::AbstractGPUMatrix, uplo::AbstractChar)
+    function LinearAlgebra.copytrito!(B::AbstractGPUMatrix, A::AbstractGPUMatrix, uplo::AbstractChar)
         LinearAlgebra.BLAS.chkuplo(uplo)
         m,n = size(A)
         m1,n1 = size(B)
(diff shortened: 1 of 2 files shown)


src/host/linalg.jl
```diff
@@ -127,6 +126,36 @@ function LinearAlgebra.copytri!(A::AbstractGPUMatrix{T}, uplo::AbstractChar, con
...
+## copy a triangular part of a matrix to another matrix
+
+if isdefined(LinearAlgebra, :copytrito!)
+    function LinearAlgebra.copytrito!(B::AbstractGPUMatrix, A::AbstractGPUMatrix, uplo::AbstractChar)
+        LinearAlgebra.BLAS.chkuplo(uplo)
+        m,n = size(A)
+        m1,n1 = size(B)
+        (m1 < m || n1 < n) && throw(DimensionMismatch("B of size ($m1,$n1) should have at least the same number of rows and columns than A of size ($m,$n)"))
+        if uplo == 'U'
+            gpu_call(A, B) do ctx, _A, _B
```
vchuravy, line 138: This looks great, but I am tempted to ask you to write it with KernelAbstractions ;)

amontoison: We already have it :)
https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/main/src/block_krylov_utils.jl#L159-L171
What is the difference between what is generated by `gpu_call` and a kernel of `KernelAbstractions.jl`?

vchuravy: `gpu_call` is the internal predecessor to KA.

The long-term plan is to move things to KA (https://github.com/JuliaGPU/GPUArrays.jl/pull/451) so this might be a good place to start.

vchuravy: Moving everything at once is a lot of work, but I am happy to merge this as is and I would be ecstatic if you found time to take on linalg

amontoison: Ok, let's do that. :smiley:

PR #533, opened 2024-05-08
Define `append!`

mtfishman:
This defines a generic version of `append!` for `AbstractGPUVector`.

It is similar to the definition for `Vector` in `Base`:
https://github.com/JuliaLang/julia/blob/6f6bb950b3e7b0a6e77bb11ffd7cc001cb87439d/base/array.jl#L1314-L1320

It relies on `resize!` and `copyto!` being defined for the `AbstractGPUVector` being appended to, while `Base` relies on `Base._growend!` and `copyto!`. Relying on `resize!` seems reasonable, most of the GPU backends, for example `CUDA.jl`, `Metal.jl`, and `AMDGPU.jl`, have `resize!` implemented. (**EDIT:** Support for `resize!` in `oneAPI.jl` was added in https://github.com/JuliaGPU/oneAPI.jl/pull/436 but it still needs to be registered).

For now it doesn't support appending with a `Tuple` of items (like `Base` allows for `Vector`) because it appears that `copyto!` doesn't generally support that for `AbstractGPUVector`s:
```julia
julia> using Metal

julia> v = mtl([1, 2, 3])
3-element MtlVector{Int64, Private}:
 1
 2
 3

julia> copyto!(v, 1, (4, 5, 6), 1, 3)
ERROR: Scalar indexing is disallowed.
Invocation of setindex! resulted in scalar indexing of a GPU array.
and therefore should be avoided.

If you want to allow scalar iteration, use `allowscalar` or `@allowscalar`
to enable scalar iteration globally or for the operations in question.
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] errorscalar(op::String)
   @ GPUArraysCore ~/.julia/packages/GPUArraysCore/GMsgk/src/GPUArraysCore.jl:155
 [3] _assertscalar(op::String, behavior::GPUArraysCore.ScalarIndexing)
   @ GPUArraysCore ~/.julia/packages/GPUArraysCore/GMsgk/src/GPUArraysCore.jl:128
 [4] assertscalar(op::String)
   @ GPUArraysCore ~/.julia/packages/GPUArraysCore/GMsgk/src/GPUArraysCore.jl:116
 [5] setindex!
   @ ~/.julia/dev/GPUArrays/src/host/indexing.jl:56 [inlined]
 [6] copyto!(dest::MtlVector{Int64, Private}, dstart::Int64, src::Tuple{Int64, Int64, Int64}, sstart::Int64, n::Int64)
   @ Base ./abstractarray.jl:1022
 [7] top-level scope
   @ REPL[16]:1
 [8] top-level scope
   @ ~/.julia/packages/Metal/q9oGt/src/initialization.jl:57

julia> using Pkg; Pkg.status("Metal")
Status `.../Metal.jl/Project.toml`
  [dde4c033] Metal v1.1.0
```
**EDIT:** The current PR now supports for appending with a Tuple by collecting it into a Vector, though perhaps it should be collected onto the GPU where the vector that is being appended lives.


tgymnich:
oneAPI.jl does not seem to have `resize!` defined.

test/testsuite/vector.jl
```diff
@@ -21,6 +21,13 @@
     # we don't shrink buffers yet... TODO shrink them... or should we?
     @test length(GPUArrays.buffer(x)) == 5
 
+    a = Float32[0, 1, 2]
+    x = AT(a)
+    append!(x, [3, 4])
+    @test length(x) == 5
+    @test x[4] == 3
+    @test x[5] == 4
+
```
tgymnich, line 30: Just a heads up. The vector tests seem to be disabled currently: https://github.com/JuliaGPU/GPUArrays.jl/blob/228fb9e8f878333cc0eda8678c8e7c2d6b5a6df2/test/testsuite.jl#L91

mtfishman:
I've addressed most comments in the latest commit:

- I re-enabled `"testsuite/vector.jl"` and deleted tests that looked outdated.
- I added support for `append!(::AbstractGPUVector, ::Tuple)`.
- I defined `resize!` for `JLArray`.

In `append!(a::AbstractGPUVector, items::Tuple)`, should the tuple `items` get allocated onto the GPU that `a` lives on?

I can work on `oneAPI.jl` support for `resize!` as best as I can but I don't have an Intel GPU available to use. Would that block this from being merged? It seems like the situation for `oneAPI.jl` before and after this PR is the same, i.e. either way right now `resize!` and `append!` don't work for `oneArray` and it seems like those could be fixed in the future.

mtfishman:
I started implementing `resize!` for `oneArray` in https://github.com/JuliaGPU/oneAPI.jl/pull/436.

src/host/abstractarray.jl
```diff
@@ -318,3 +318,19 @@ Base.deepcopy(x::AbstractGPUArray) = copy(x)
...
+
+function Base.append!(a::AbstractGPUVector, items::AbstractVector)
+    n = length(items)
+    resize!(a, length(a) + n)
+    copyto!(a, length(a) - n + 1, items, firstindex(items), n)
+    return a
+end
+
+# this is needed because copyto! of most GPU arrays
+# doesn't support Tuple sources write now
```
mtfishman, line 332:
```suggestion
# doesn't currently support Tuple sources
```

PR #538, opened 2024-05-30
Accomodate for rectangular matrices in copytrito!

amontoison:
Related to https://github.com/JuliaLang/julia/issues/54546 and https://github.com/JuliaLang/julia/pull/54587.
cc @dkarrasch


dkarrasch (approved):
Do you need tests for the failing cases?

PR #550, opened 2024-07-06
Introduce `AsyncNumber` to lazily copy numeric `mapreduce` results to the host

pxl-th:
Introduce `GPUNumber` to store the resul of `mapreduce` across all `dims` (i.e. `dims = :`) instead of immediately transferring it to host.

`GPUNumber` copies its value to host when it is used as a regular number.
But if it is used for broadcasting, it passes its underlying 1-element array without device-to-host copy.

Seems to be minimally breaking change, except for the return type of `sum` and the likes.
Using AbstractNumbers.jl since it implements the whole `Number` interface conveniently.

More context: https://github.com/FluxML/Zygote.jl/issues/1513

- Before:

```julia
julia> x = AMDGPU.rand(Float32, 1024);

julia> @btime sum($x);
  64.340 μs (146 allocations: 6.11 KiB)

julia> @btime Zygote.gradient($x) do x
           sum(1f0 .- x)
       end;
  131.617 μs (368 allocations: 16.10 KiB)
```

- Now:

```julia
julia> @btime sum($x);
  15.809 μs (140 allocations: 5.84 KiB)

julia> @btime Zygote.gradient($x) do x
           sum(1f0 .- x)
       end;
  44.123 μs (356 allocations: 15.57 KiB)
```

Here's also a timeline profile for the Flux.jl model for the same region.
We can see that now there are no device-to-host copies.

||Timeline profile|
|-|-|
|Before|image|
|Now|image|

### DIFF
--- a/src/host/gpunumber.jl
+++ b/src/host/gpunumber.jl
@@ -38,3 +38,5 @@ Base.isequal(v::Number, g::GPUNumber) = isequal(v, g[])
 Base.nextpow(a, x::GPUNumber) = nextpow(a, x[])
 Base.nextpow(a::GPUNumber, x) = nextpow(a[], x)
 Base.nextpow(a::GPUNumber, x::GPUNumber) = nextpow(a[], x[])
+
+Base.convert(::Type{Number}, g::GPUNumber) = g[]
--- a/src/host/gpunumber.jl
+++ b/src/host/gpunumber.jl
@@ -26,5 +26,7 @@ Base.broadcastable(g::GPUNumber) = g.val
 
 # Overload to avoid copies.
 Base.one(g::GPUNumber) = one(number_type(g))
+Base.one(::Type{GPUNumber{T}}) where T = one(eltype(T))
 Base.zero(g::GPUNumber) = zero(number_type(g))
+Base.zero(::Type{GPUNumber{T}}) where T = zero(eltype(T))
 Base.identity(g::GPUNumber) = g
--- a/src/host/gpunumber.jl
+++ b/src/host/gpunumber.jl
@@ -1,0 +1,30 @@
+# Custom GPU-compatible `Number` interface.
+struct GPUNumber{T <: AbstractGPUArray} <: AN.AbstractNumber{T}
+    val::T
+
+    function GPUNumber(val::T) where T <: AbstractGPUArray
+        length(val) != 1 && error(
+            "`GPUNumber` accepts only 1-element GPU arrays, " *
+            "instead `$(length(val))`-element array was given.")
+        new{T}(val)
+    end
+end
+
+AN.number(g::GPUNumber) = @allowscalar g.val[]
+
+maybe_number(g::GPUNumber) = AN.number(g)
+maybe_number(g) = g
+
+number_type(::GPUNumber{T}) where T = eltype(T)
+
+# When operations involve other `::Number` types,
+# do not convert back to `GPUNumber`.
+AN.like(::Type{<: GPUNumber}, x) = x
+
+# When broadcasting, just pass the array itself.
+Base.broadcastable(g::GPUNumber) = g.val
+
+# Overload to avoid copies.
+Base.one(g::GPUNumber) = one(number_type(g))
+Base.zero(g::GPUNumber) = zero(number_type(g))
+Base.identity(g::GPUNumber) = g
--- a/src/host/gpunumber.jl
+++ b/src/host/gpunumber.jl
@@ -11,7 +11,6 @@ struct GPUNumber{T <: AbstractGPUArray} <: AN.AbstractNumber{T}
 end
 
 AN.number(g::GPUNumber) = @allowscalar g.val[]
-
 maybe_number(g::GPUNumber) = AN.number(g)
 maybe_number(g) = g
 
@@ -30,3 +29,12 @@ Base.one(::Type{GPUNumber{T}}) where T = one(eltype(T))
 Base.zero(g::GPUNumber) = zero(number_type(g))
 Base.zero(::Type{GPUNumber{T}}) where T = zero(eltype(T))
 Base.identity(g::GPUNumber) = g
+
+Base.getindex(g::GPUNumber) = AN.number(g)
+
+Base.isequal(g::GPUNumber, v::Number) = isequal(g[], v)
+Base.isequal(v::Number, g::GPUNumber) = isequal(v, g[])
+
+Base.nextpow(a, x::GPUNumber) = nextpow(a, x[])
+Base.nextpow(a::GPUNumber, x) = nextpow(a[], x)
+Base.nextpow(a::GPUNumber, x::GPUNumber) = nextpow(a[], x[])
--- a/src/host/gpunumber.jl
+++ b/src/host/gpunumber.jl
@@ -1,42 +1,42 @@
 # Custom GPU-compatible `Number` interface.
-struct GPUNumber{T <: AbstractGPUArray} <: AN.AbstractNumber{T}
+struct AsyncNumber{T <: AbstractGPUArray} <: AbstractNumbers.AbstractNumber{T}
     val::T
 
-    function GPUNumber(val::T) where T <: AbstractGPUArray
+    function AsyncNumber(val::T) where T <: AbstractGPUArray
         length(val) != 1 && error(
-            "`GPUNumber` accepts only 1-element GPU arrays, " *
+            "`AsyncNumber` accepts only 1-element GPU arrays, " *
             "instead `$(length(val))`-element array was given.")
         new{T}(val)
     end
 end
 
-AN.number(g::GPUNumber) = @allowscalar g.val[]
-maybe_number(g::GPUNumber) = AN.number(g)
+AbstractNumbers.number(g::AsyncNumber) = @allowscalar g.val[]
+maybe_number(g::AsyncNumber) = AbstractNumbers.number(g)
 maybe_number(g) = g
 
-number_type(::GPUNumber{T}) where T = eltype(T)
+number_type(::AsyncNumber{T}) where T = eltype(T)
 
 # When operations involve other `::Number` types,
-# do not convert back to `GPUNumber`.
-AN.like(::Type{<: GPUNumber}, x) = x
+# do not convert back to `AsyncNumber`.
+AbstractNumbers.like(::Type{<: AsyncNumber}, x) = x
 
 # When broadcasting, just pass the array itself.
-Base.broadcastable(g::GPUNumber) = g.val
+Base.broadcastable(g::AsyncNumber) = g.val
 
 # Overload to avoid copies.
-Base.one(g::GPUNumber) = one(number_type(g))
-Base.one(::Type{GPUNumber{T}}) where T = one(eltype(T))
-Base.zero(g::GPUNumber) = zero(number_type(g))
-Base.zero(::Type{GPUNumber{T}}) where T = zero(eltype(T))
-Base.identity(g::GPUNumber) = g
+Base.one(g::AsyncNumber) = one(number_type(g))
+Base.one(::Type{AsyncNumber{T}}) where T = one(eltype(T))
+Base.zero(g::AsyncNumber) = zero(number_type(g))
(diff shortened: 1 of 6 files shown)


pxl-th:
@maleadt, the remaining CUDA.jl issues are because of method signatures, which can easily be updated to handle the change. I can open respective PR in CUDA.jl if you think this change is fine

maleadt:
Surprisingly small change! Still not a fan though :-P How much additional pressure does this put on the GC?

> which can easily be updated to handle the change

Only without losing some "type safety", right? IIUC `<:Integer` arguments have to be widened to `<:Number`, or similarly `<:Real` or `<:Complex` types would get lost.

pxl-th:
To avoid changing method signatures we can just change [this](https://github.com/JuliaGPU/CUDA.jl/blob/229d13f88fece1bd1dd6422575d61edf1e0cb753/lib/cusparse/conversions.jl#L35) line to:
```diff
- m=maximum(I), n=maximum(J);
+ m=maximum(I)[], n=maximum(J)[];
```

> IIUC <:Integer arguments have to be widened to <:Number, or similarly <:Real or <:Complex types would get lost.

You mean that the compiler wouldn't be able to infer the type?
I think it is type-stable, since we maintain the underlying type in `GPUNumber{T}` and `GPUNumber` just forwards through a thin layer.

```julia
julia> x = AMDGPU.rand(Float32, 16);

julia> f(x::Number) = x + 2
f (generic function with 1 method)

julia> @code_warntype sum(x)
MethodInstance for sum(::ROCArray{Float32, 1, AMDGPU.Runtime.Mem.HIPBuffer})
  from sum(a::AbstractArray; dims, kw...) @ Base reducedim.jl:1010
Arguments
  #self#::Core.Const(sum)
  a::ROCArray{Float32, 1, AMDGPU.Runtime.Mem.HIPBuffer}
Body::GPUArrays.GPUNumber{ROCArray{Float32, 1, AMDGPU.Runtime.Mem.HIPBuffer}}
1 ─      nothing
│   %2 = Base.:(:)::Core.Const(Colon())
│   %3 = Core.NamedTuple()::Core.Const(NamedTuple())
│   %4 = Base.pairs(%3)::Core.Const(Base.Pairs{Symbol, Union{}, Tuple{}, @NamedTuple{}}())
│   %5 = Base.:(var"#sum#828")(%2, %4, #self#, a)::GPUArrays.GPUNumber{ROCArray{Float32, 1, AMDGPU.Runtime.Mem.HIPBuffer}}
└──      return %5

julia> @code_warntype(f(sum(x)))
MethodInstance for f(::GPUArrays.GPUNumber{ROCArray{Float32, 1, AMDGPU.Runtime.Mem.HIPBuffer}})
  from f(x::Number) @ Main REPL[3]:1
Arguments
  #self#::Core.Const(f)
  x::GPUArrays.GPUNumber{ROCArray{Float32, 1, AMDGPU.Runtime.Mem.HIPBuffer}}
Body::Float32
1 ─ %1 = (x + 2)::Float32
└──      return %1
```

pxl-th:
> How much additional pressure does this put on the GC?

For the Flux model that I have and use for testing, machine consistently hangs (machine with a single AMD GPU) after several hundred training steps with or without this change...
So I'm not sure it should be a blocker for this PR.

But generally, I really do think that something needs to be done to GC and GPU, since this is the biggest blocker for Julia + DL right now, unless you have 2+ GPUs or use headless server...

maleadt:
> > IIUC <:Integer arguments have to be widened to <:Number, or similarly <:Real or <:Complex types would get lost.
> 
> You mean that the compiler wouldn't be able to infer the type?

I'm not thinking of type stability or inference, but the ability to restrict method definitions (e.g. for dispatch purposes). If we now have `foo(::Real)` and `foo(::Complex)` we lose the ability to distinguish between both by doing `foo(::GPUNumber)`. Or when you have a method `bar(::CuArray{T}, ::T) where {T <: Number}` signalling that the (el)types need to match; this would become impossible too.

The fact that we're now indicating that `mapreduce` returns a `Number` seems wrong to me too. It's possible to reduce non-numbers, e.g., https://github.com/JuliaGPU/GPUArrays.jl/blob/80b82267d15a9ae10d90228bb3899b2be68d996e/src/host/mapreduce.jl#L111-L134. I take it this works because `getproperty` is forwarded to the inner value? That doesn't seem workable though, as you can't forward _everything_ to the inner value (and with arbitrary values being possible here, it's impossible to tell what to forward).

pxl-th:
> I take it this works because getproperty is forwarded to the inner value?

No, `GPUNumber` only inherits `Number` interface. For everything else (like with that `reducer` example) the user would need to explicitly tranfer value to host by calling getindex `sum(x)[]` (or `AN.number` as is in this PR).

> The fact that we're now indicating that mapreduce returns a Number also seems wrong to me too.

Maybe we can have `GPUNumber` for numbers and some other container for non-`Number` types?

>  Or when you have a method bar(::CuArray{T}, ::T) where {T <: Number} signalling that the (el)types need to match; this would become impossible too.

Yeah, that's a consequence of making host transfers more explicit I guess...
Maybe we can expose something like (as defined in this PR):
```julia
maybe_number(g::GPUNumber) = AN.number(g)
maybe_number(g) = g
```

So that if the user writes code both for CPU and GPU he can use `maybe_number` (or some other name, like `materialize`) when needed to make sure the type is actually a scalar.

src/host/gpunumber.jl
```diff
@@ -0,0 +1,42 @@
+# Custom GPU-compatible `Number` interface.
+struct GPUNumber{T <: AbstractGPUArray} <: AN.AbstractNumber{T}
+    val::T
+
+    function GPUNumber(val::T) where T <: AbstractGPUArray
```
vpuri3, line 5: @pxl-th it might be a good idea to do a `dropdims` or `vec` in the constructor so this guy truly behaves like a number. The case I am thinking of is the following.

```julia
a = CUDA.rand(1,1,1,1) |> GPUNumber
x = CUDA.rand(4,4)
size(x .+ a) # want (4,4)
```

I haven't run your code but it seems like it would do something similar to
```julia
julia> CUDA.rand(1,1,1,1) .+ CUDA.rand(4,4) |> size
(4, 4, 1, 1)
```

maleadt:
> For everything else (like with that `reducer` example) the user would need to explicitly tranfer value to host by calling getindex `sum(x)[]`

But that's a breaking change? So yeah, making this happen only for Numbers where we can be "sure" we implement the number interface transparently seems like the safer choice. But even then I'm still wary that this will break user code, so this will need thorough testing...

src/host/gpunumber.jl
pxl-th: Indeed, thanks!

pxl-th:
So I've made it behave as usual when `eltype` is not `Number`, otherwise return `GPUNumber`.
I'll also do some more testing and benchmarking to see the impact.

src/host/gpunumber.jl
```diff
@@ -0,0 +1,42 @@
+# Custom GPU-compatible `Number` interface.
+struct GPUNumber{T <: AbstractGPUArray} <: AN.AbstractNumber{T}
```
vchuravy, line 2: I prefer this being written as `AbstractNumbers.`

Then I don't have to remember what `AN` stands for when reading the code.

maleadt:
Also, some bike shedding: `AsyncNumber` might be more clear than the otherwise noninformative `GPUNumber`. Or `AsyncValue` if we find a way to generalize this, which I'm not sure we will. Also sounds a bit like a `Ref`, but that verb is pretty overloaded already.

pxl-th:
Addressed comments.

Here's also a timeline view for 20 training steps of a Flux model using CUDA.jl.
Before there are a ton of host synchronizations and a total time is ~29 seconds.
With this PR and couple of others (that also remove device-host transfers) the time is ~19 seconds and there are no device-host transfers:
- https://github.com/JuliaDiff/ChainRules.jl/pull/801
- https://github.com/JuliaGPU/GPUArrays.jl/pull/548 (although it still does bounds checking in ChainRules rules, so I had to manually comment it out).

The profiled code explicitly avoids displaying loss values, so in real training where you'd log the model performance they will be, but without logging there should be no such transfers.

That said, we are still ~3x slower that equivalent PyTorch code on CUDA, where each epoch is taking ~12 minutes, but in Julia it is ~35-40 minutes...
So needs further investigation.

||Timeline|
|-|-|
|Before|image|
|Now|image|

pxl-th:
Here's also a timeline for the Flux.jl model for CUDA.jl.
Profiling over 20 training steps and explicitly avoiding any host transfers, like visualizing loss values.
Before it took ~29 seconds, now ~19 seconds. That includes some other PRs that avoid host transfers:
- https://github.com/JuliaDiff/ChainRules.jl/pull/801
- https://github.com/JuliaGPU/GPUArrays.jl/issues/542 (although it still does bounds checking in ChainRules.jl so I disabled it manually)

When running outside of profiler, one epoch takes ~35-40 minutes (down from 50-55 min), while PyTorch takes ~12 minutes per epoch. So there's still room for improvement (timeline profiling also looks much denser for PyTorch, there are no gaps between GPU kernels).

||Timeline|
|-|-|
|Before|image|
|Now|image|
|PyTorch|image|

pxl-th:
Remaining gaps could be either GC pauses.
Running profiling with GC logging enabled:
```julia
GC: pause 366.60ms. collected 5.253326MB. incr 

GC: pause 112.91ms. collected 34.399342MB. full recollect

GC: pause 355.50ms. collected 0.455795MB. incr 

GC: pause 114.99ms. collected 30.454090MB. full recollect

GC: pause 371.72ms. collected 0.405960MB. incr 

GC: pause 118.76ms. collected 30.621845MB. full recollect

GC: pause 359.00ms. collected 5.808136MB. incr 

GC: pause 124.97ms. collected 10.381500MB. full recollect

GC: pause 348.96ms. collected 5.253326MB. incr 

GC: pause 112.76ms. collected 37.985279MB. full recollect

GC: pause 348.96ms. collected 0.455795MB. incr 

GC: pause 111.61ms. collected 33.037785MB. full recollect

GC: pause 349.60ms. collected 0.405960MB. incr 
```

Or synchronizations during host -> device transfer. Could be that PyTorch's data loader prefetches data on a separate worket/stream to overlap with compute (don't think it is using pinned memory by default).

maleadt:
You could use NVTX.jl to visualize GC pauses, https://juliagpu.github.io/NVTX.jl/dev/#Instrumenting-Julia-internals, and maybe add a couple of annotations to key CUDA.jl functions.

pxl-th:
> You could use NVTX.jl to visualize GC pauses

Screenshot from 2024-07-12 16-16-25

So it does look like these gaps are GC pauses. I've also logged `maybe_collect` and it almost always corresponds to one of these GC pauses and the pressure when it is triggered is always `> 0.99`.

maleadt:
Is the GC time spent marking/sweeping in Julia, or are the `cuMemFreeAsync` calls soaking up the time? I would assume the former, but I've also seen CUDA itself spending significant time in the allocator especially when close to OOM (presumably compacting/defragmenting its heap).

vchuravy:
> So it does look like these gaps are GC pauses.

Can you set `--gcthreads` to be 4 or so?

vpuri3:
@pxl-th what model are you training? I'm curious how much of the performance gain is due to AsyncNumber vs https://github.com/JuliaDiff/ChainRules.jl/pull/801

pxl-th:
It was [HiFi-GAN](https://arxiv.org/abs/2010.05646) and https://github.com/JuliaDiff/ChainRules.jl/pull/801 probably brings bigger performance gain.
And I'm not sure at this point how impactful this PR is in the real-world use-cases, because in the end the real bottleneck was due to GC calls when we ran out of memory as described above https://github.com/JuliaGPU/GPUArrays.jl/pull/550#issuecomment-2225915738.

maleadt:
Why the close? Did this turn out to hurt GC too much?

pxl-th:
Ah... I accidentally removed my fork of `GPUArrays` and forgot it has this PR...
We should reopen it

maleadt:
I've added you to the AMDGPU team and give you access to GPUArrays.jl, so you can just host your branch here if you want.

PR #590, opened 2025-04-13
Faster fallback matrix multiplication

christiangnrd:
Adapted from the Metal [scalar gemm kernel](https://github.com/JuliaGPU/Metal.jl/blob/main/src/gemm.jl) and modified for the macro-based KernelAbstractions interface based on the KA "performant_matmul.jl" example but with some fixes.


maleadt:
> I had to make a few changes, such as using `unsafe_indices`, since the algorithm itself does the bounds checking, and I was getting wrong results until I added that.

Oof, that's bad, and unexpected. cc @vchuravy 

> Guess i found out why that was there.

Care to elaborate?

Any performance numbers?

christiangnrd:
> > I had to make a few changes, such as using `unsafe_indices`, since the algorithm itself does the bounds checking, and I was getting wrong results until I added that.
> 
> Oof, that's bad, and unexpected. cc @vchuravy

To reproduce, you can apply this patch:
<details>
<summary>Patch</summary>

```diff
diff --git a/src/host/linalg.jl b/src/host/linalg.jl
index b59598f..2e51d9f 100644
--- a/src/host/linalg.jl
+++ b/src/host/linalg.jl
@@ -326,7 +326,7 @@ function LinearAlgebra.ldiv!(B::AbstractGPUVecOrMat,
 end

 # XXX: figure out how to do dynamically
-MAX_TILE_DIM = 16
+MAX_TILE_DIM = 2 # THIS CHANGE MADE TO SIMPLIFY MWE OUTPUT

 ## matrix multiplication
 # legacy method
@@ -346,7 +346,7 @@ function generic_matmatmul!(C::AbstractGPUMatrix{R}, A::AbstractGPUMatrix{T}, B:
         return fill!(C, zero(R))
     end

-    @kernel unsafe_indices=true function coalesced_matmul_kernel!(
+    @kernel function coalesced_matmul_kernel!(
             output, @Const(input1), @Const(input2), N, Q, M,
             ::Val{BANK} = Val(1),
         ) where {BANK}
@@ -408,7 +408,7 @@ function generic_matmatmul!(C::AbstractGPUMatrix{R}, A::AbstractGPUMatrix{T}, B:
         end
     end

-    coalesced_matmul_kernel!(get_backend(C), (MAX_TILE_DIM, MAX_TILE_DIM))(C, A, B, N, Q, M;ndrange=map(x -> ceil(Int,x/MAX_TILE_DIM)*MAX_TILE_DIM, size(C)))
+    coalesced_matmul_kernel!(get_backend(C), (MAX_TILE_DIM, MAX_TILE_DIM))(C, A, B, N, Q, M;ndrange=size(C))
     C
 end
 function generic_matmatmul!(C::AbstractArray{R}, A::AbstractArray{T}, B::AbstractArray{S}, add::MulAddMul) where {T,S,R}
```

</details>

And then when run using Metal/CUDA, it gives:
```julia-repl
julia> using Metal, GPUArrays; a = Metal.ones(3,3); b = Metal.ones(3,3); c = Metal.zeros(3,3); GPUArrays.generic_matmatmul!(c,a,b, true, false)
Precompiling Metal...
  3 dependencies successfully precompiled in 13 seconds. 66 already precompiled.
3×3 MtlMatrix{Float32, Metal.PrivateStorage}:
 3.0  3.0   2.0
 3.0  3.0   2.0
 2.0  2.0  16.6155

# With CUDA
julia> using CUDA, GPUArrays; a = CUDA.ones(3,3); b = CUDA.ones(3,3); c = CUDA.zeros(3,3); GPUArrays.generic_matmatmul!(c,a,b, true, false)
Precompiling CUDA...
  3 dependencies successfully precompiled in 38 seconds. 96 already precompiled.
3×3 CuArray{Float32, 2, CUDA.DeviceMemory}:
 3.0  3.0  2.0
 3.0  3.0  2.0
 2.0  2.0  2.0
```

It seems to not be broken with JLArrays, and with CUDA it seems less broken in that it results in close-to-integer-valued results in the other 3 quadrants. This happens on KA 0.9.34, I haven't tested with master branch.

> > Guess i found out why that was there.
> 
> Care to elaborate?

Yes sorry about that. CI was showing that on some platforms, `I` and `J` were no longer in scope/defined after an `@synchronize` call. Without looking into it for this case specifically, I assume it has to do with the code transformations the `@kernel` macro does.

> Any performance numbers?

It seems to be at least as fast and up to 4-5x faster than the naive algorithm. 

Linux Ryzen 3700X with RTX 3060 (note the different y-axes in the bottom row):
bench_all_1_3060

M2 Max 30 core:
bench_all_1

christiangnrd:
Based on https://github.com/JuliaGPU/KernelAbstractions.jl/pull/590 passing tests, maybe this should wait until GPUArrays supports KA v0.10

christiangnrd:
OpenCL POCL before and after:

| eltype | size | main | PR | speedup |
|--------|------|-----:|---:|--------:|
| Float16 | 512³  | 23 | 45 | 2.0× |
| Float16 | 1024³ | 22 | 47 | 2.1× |
| Float16 | 2048³ | 13 | 45 | 3.5× |
| Float16 | 4096³ | 2  | 45 | 22.5× |
| Int32   | 512³  | 24 | 60 | 2.5× |
| Int32   | 1024³ | 13 | 60 | 4.6× |
| Int32   | 2048³ | 4  | 59 | 14.8× |
| Int32   | 4096³ | 2  | 37 | 18.5× |
| Float32 | 512³  | 21 | 44 | 2.1× |
| Float32 | 1024³ | 13 | 46 | 3.5× |
| Float32 | 2048³ | 4  | 44 | 11.0× |
| Float32 | 4096³ | 2  | 31 | 15.5× |

OpenCL RUSTICL before and after:

| eltype | size | main | PR | speedup |
|--------|------|-----:|---:|--------:|
| Float16 | 512³  | 159 | 357 | 2.0× |
| Float16 | 1024³ | 214 | 476 | 2.1× |
| Float16 | 2048³ | 230 | 508 | 3.5× |
| Float16 | 4096³ | 212  | 490 | 22.5× |
| Int32   | 512³  | 151 | 329 | 2.5× |
| Int32   | 1024³ | 210 | 455 | 4.6× |
| Int32   | 2048³ | 216  | 472 | 14.8× |
| Int32   | 4096³ | 194 | 432 | 18.5× |
| Float32 | 512³  | 151 | 330 | 2.1× |
| Float32 | 1024³ | 208 | 457 | 3.5× |
| Float32 | 2048³ | 214 | 472 | 11.0× |
| Float32 | 4096³ | 194 | 331 | 15.5× |

christiangnrd:
I updated this to include the overflow and other various improvements recently applied to the fallback matmul, and reimplemented on top of KernelAbstractions instead of KernelInterface so we can merge sooner rather than later.

maleadt:
> > I had to make a few changes, such as using `unsafe_indices`, since the algorithm itself does the bounds checking, and I was getting wrong results until I added that.
> 
> Oof, that's bad, and unexpected.

I had Claude look into the `unsafe_indices` question. It's not a bug in KA or Metal — it's KA's index masking doing more than one might expect.

Without `unsafe_indices`, `@kernel` splits the body at every `@synchronize` and wraps *each chunk of user code* in `if __validindex(ctx)`, leaving only the barriers unconditional. So when `ndrange = size(C)` isn't a multiple of the tile, out-of-range lanes still hit the barriers (good), but they never execute their `tile[li, lj] = ...` loads. In-range lanes then read stale threadgroup memory for those slots — `3 3 2 / 3 3 2 / 2 2 2` is one real product plus one stale zero per tile, and the `16.6` on Metal is the same slot containing garbage (Metal doesn't zero local memory; CUDA happened to).

Confirmed on Metal with TILE=2, 3×3: masked kernel + `ndrange=(3,3)` is wrong; masked + padded `ndrange=(4,4)` is fine; `unsafe_indices=true` is fine either way.

So for a cooperative tiled kernel that does its own bounds checks, `unsafe_indices=true` is the right declaration — all lanes have to participate in the loads. Padding the ndrange alone would also work, but I'd keep the flag (with a comment) so it doesn't silently regress. The `I`/`J`-undefined-after-`@synchronize` issue is the same split machinery: each chunk is its own scope, and only the CPU path still splits when `unsafe_indices=true`, which is why it showed up on some backends only.

Note that KA's `performant_matmul.jl` example has the same latent problem (`ndrange = size(C)`, no `unsafe_indices`); it only works for tile-aligned sizes.

christiangnrd:
> Note that KA's `performant_matmul.jl` example has the same latent problem (`ndrange = size(C)`, no `unsafe_indices`); it only works for tile-aligned sizes.

The 0.9 example hasn't been fixed, but the 0.10 example got fixed (and then I switched it over to KernelInterface so it was no longer an issue)

I've updated the PR description to reflect the current state

maleadt:
Pushed a handful of clean-ups/minimizations by Fable.

christiangnrd:
Nice! Should we stack this on top of https://github.com/JuliaGPU/GPUArrays.jl/pull/772 to make sure those tests pass before merging?

Also, do you mind applying the https://github.com/JuliaGPU/GPUArrays.jl/pull/590/commits/018ea2f44343c98bc7c8c2fda4b4b28e753a0a80 change in Metal? That code was ripped from there

maleadt:
> Nice! Should we stack this on top of #772 to make sure those tests pass before merging?

Yeah that's reasonable.

> Also, do you mind applying the [018ea2f](https://github.com/JuliaGPU/GPUArrays.jl/commit/018ea2f44343c98bc7c8c2fda4b4b28e753a0a80) change in Metal? That code was ripped from there

https://github.com/JuliaGPU/Metal.jl/pull/935

christiangnrd:
Julia 1.11 Metal [failure](https://buildkite.com/julialang/gpuarrays-dot-jl/builds/1910/table?jid=01a03e7c-6eaf-48f5-95c0-9db146213bfe&tab=output) seems unrelated?

christiangnrd:
This PR on from my fork lets just wait for #772 to be merged and rebase.

maleadt:
CI failure is https://github.com/JuliaGPU/Metal.jl/issues/937

maleadt (approved):

maleadt:
> This PR on from my fork lets just wait for #772 to be merged and rebase.

OK, I'll let you apply the stack if you don't mind.

christiangnrd:
> OK, I'll let you apply the stack if you don't mind.

https://github.com/JuliaGPU/GPUArrays.jl/pull/773

PR #592, opened 2025-05-13
KernelAbstractions 0.10 compat

christiangnrd:
For testing

PR #593, opened 2025-06-06
Fix version bounds for broken test

christiangnrd:
Found while working on #592. This should be much quicker to merge on its own


christiangnrd:
This removes the one error I expected it to (in Array/base) and doesn't add any new errors so it's ready for review

maleadt (approved):

PR #597, opened 2025-06-16
JLD2 extension take 2

christiangnrd:
@lassepe I accidentally messed up your PR (#596) so I'm reopening this. I'll try to make it pass tests but if you get it fully working before then feel free to reopen and I'll close this one.

I apologize for the inconvenience.


christiangnrd:
I got the tests to pass. However, I'm not sure this is the best way to deal with the new JLD2 requirement. Do we add a JLD2 dependency directly to this package's setup file so that the downstream packages require no changes, or do we add JLD2 to every downstream package's test Project.toml?

lassepe:
Friendly bump before this goes stale, @christiangnrd :)

PR #604, opened 2025-07-14
Test a few more corner cases in reductions

christiangnrd:
Adds a few corner case tests for reductions based on the AK.jl reduction tests.


christiangnrd:
Test failures are either expected or fixed with https://github.com/JuliaGPU/GPUArrays.jl/pull/605.

PR #630, opened 2025-10-09
Try to unbreak linalg test on nightly

kshyatt:
This is causing CUDA nightly to fail because we don't have a specialized method to copy triangular types on GPU to CPU matrices. This is a work around that hopefully avoids the fail while still test appropriately.

### DIFF
--- a/test/testsuite/linalg.jl
+++ b/test/testsuite/linalg.jl
@@ -124,8 +124,8 @@
                 gpu_b = AT{Float32}(undef, 128, 128) |> TR
 
                 gpu_c = copyto!(gpu_b, gpu_a)
-                @test all(Array(gpu_b) .== Array(gpu_a))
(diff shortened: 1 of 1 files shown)


christiangnrd:
@maleadt https://github.com/JuliaLang/LinearAlgebra.jl/pull/1263 and https://github.com/JuliaLang/LinearAlgebra.jl/pull/1282 have made it so that `copytrito!` functions are bypassed.

Since I don't believe the broken code is actually what's being tested, imo this PR makes the testsuite more robust.

Unless you think it'll get forgotten unless fixed by this PR, I think we should merge this and then open an issue for supporting and testing`Matrix{T}(U::UpperOrLowerTriangular{T, S<:AbstractGPUMatrix)`.

Broken since https://github.com/JuliaLang/julia/pull/58180 or https://github.com/JuliaLang/julia/pull/58396

PR #642, opened 2025-10-28
edge case: `tril!`, `triu!` and `I` constructors with empty matrices

tam724:
Shortcuts the kernels in `triu!`, `tril!` and `identity_kernel` if the arguments are empty matrices.

All three would fail with `DivideError: integer division error` (from KernelAbstractions.jl).

### DIFF
--- a/test/testsuite/linalg.jl
+++ b/test/testsuite/linalg.jl
@@ -310,10 +310,10 @@
         @testset "$f with diagonal $d" for f in (triu, triu!, tril, tril!),
                                             d in -2:2
             A = randn(Float32, 10, 10)
-            @test compare(f, AT, A)
+            @test compare(f, AT, A, d)
 
             A_empty = randn(Float32, 0, 0)
-            @test compare(f, AT, A_empty)
+            @test compare(f, AT, A_empty, d)
         end
     end
 


src/host/construction.jl
```diff
@@ -48,9 +49,10 @@ end
 
 function Base.copyto!(A::AbstractGPUMatrix{T}, s::UniformScaling) where T
     fill!(A, zero(T))
+    if iszero(minimum(size(A))) return A end
```
maleadt:
```suggestion
   iszero(minimum(size(A))) && return A
```

src/host/linalg.jl
```diff
@@ -165,27 +165,29 @@ for T in (UpperTriangular, LowerTriangular, UnitUpperTriangular, UnitLowerTriang
 end
 
 function LinearAlgebra.tril!(A::AbstractGPUMatrix{T}, d::Integer = 0) where T
-  @kernel function tril_kernel!(_A, _d)
-    I = @index(Global, Cartesian)
-    i, j = Tuple(I)
-    if i < j - _d
-      @inbounds _A[i, j] = zero(T)
+    if isempty(A) return A end
```
maleadt:
```suggestion
    isempty(A) && return A
```

test/testsuite/linalg.jl
```diff
@@ -307,10 +307,13 @@
             @test_throws SingularException ldiv!(D, B)
         end
 
-        @testset "$f! with diagonal $d" for (f, f!) in ((triu, triu!), (tril, tril!)),
+        @testset "$f with diagonal $d" for f in (triu, triu!, tril, tril!),
                                             d in -2:2
             A = randn(Float32, 10, 10)
-            @test f(A, d) == Array(f!(AT(A), d))
+            @test compare(f, AT, A)
```
christiangnrd:
```suggestion
            @test compare(f, AT, A, d)
```

maleadt (approved):

PR #646, opened 2025-11-04
add zero dim matmat/matvec-mul tests

tam724:
see https://github.com/JuliaGPU/CUDA.jl/pull/2958.


tam724:
Test failures are expected:
CUDA.jl requires https://github.com/JuliaGPU/CUDA.jl/pull/2958 to pass. Metal.jl and oneAPI.jl need similar changes.

PR #647, opened 2025-11-12
Document all public symbols

gdalle:
I did my best to add a docstring to all public symbols and include them in the documentation website. There were a few for which I didn't know what to say

### DIFF
--- a/lib/GPUArraysCore/src/GPUArraysCore.jl
+++ b/lib/GPUArraysCore/src/GPUArraysCore.jl
@@ -13,7 +13,7 @@ export AbstractGPUArray, AbstractGPUVector, AbstractGPUMatrix, AbstractGPUVecOrM
     AbstractGPUArray{T, N} <: DenseArray{T, N}
 
 Supertype for `N`-dimensional GPU arrays (or array-like types) with elements of type `T`.
-Instances of this type are expected to live on the host, see [`AbstractDeviceArray`](@ref)
+Instances of this type are expected to live on the host, see `AbstractDeviceArray`
 for device-side objects.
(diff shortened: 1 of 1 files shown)


lib/JLArrays/src/JLArrays.jl
```diff
@@ -123,6 +139,18 @@ mutable struct JLArray{T, N} <: AbstractGPUArray{T, N}
...
+    JLSparseVector{Tv, Ti}
+
+Sparse vector backed by `JLVector`s, similar to `SparseArrays.SparseVector`.
+
+# Fields
+
+- `iPtr::JLVector{Ti, 1}`: indices of non-zero coefficients
+- `nzVal::JLVector{Tv, 1}`: values of non-zero coefficients
+- `len::Int`: size of the vector
+- `nnz::Ti`: number of non-zero coefficients
```
maleadt: I'm personally not a fan of unconditionally documenting all fields, especially when they are implicitly set by constructors and exposed through interface methods (nnzvals, length, etc). They aren't public API, but sadly there's no way to express that.

I'd be fine doing this for some other structures, so it's not a general rule though.

gdalle: > They aren't public API, but sadly there's no way to express that.

That's why I want to be careful about this PR, because in Julia 1.10 documented = public by default. In Julia 1.11 though we could add the public keyword!

gdalle:
How do I avoid running the full test suite for this PR? Is there a mechanism similar to CUDA's?

gdalle:
Superseded by #653

PR #660, opened 2025-11-24
Return early (don't kernel launch) for copytrito and zero-sized arrays

kshyatt:
This was getting hit at https://buildkite.com/julialang/cuda-dot-jl/builds/6739#019ab523-924c-4308-8219-1305d58116da. Not sure why the existing buildkite tests didn't detect it.


christiangnrd (approved):
I tested this PR (rebased) and it fixes the error on Metal

PR #662, opened 2025-11-26
Make stats tests numerically more stable for oneAPI

michel2323:
On oneAPI.jl, the cor(A; dims=2) test failed for `Float16` (one some GPUs), caused by numerical instability when computing the variance of rows with only 2 elements drawn from [0, 1). This leads to subnormal variances, which are flushed to zero on the CPU (causing NaN or -1 results) but preserved on the oneAPI GPU (causing valid results), resulting in a mismatch. 

Now I'm not a stats person, and I don't want to fix anything deeper in Intel oneAPI, but it seems to be fixed with an extended range. I saw that there was already such a special case for `Real`.  Unfortunately, that still doesn't fix the `ComplexF32` case.

### DIFF
--- a/test/testsuite/statistics.jl
+++ b/test/testsuite/statistics.jl
@@ -5,7 +5,8 @@ using Statistics
         if !(ET in [Float16, Float32, Float64])
             continue
         end
-        range = ET <: Real ? (ET(1):ET(100)) : ET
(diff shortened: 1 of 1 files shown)


michel2323:
> Conceptually LGTM, but it still looks somewhat sketchy to do this for e.g. Float64...

I limited the extended range to Float16.

PR #663, opened 2025-11-28
Move `kron` and `diagonal` tests out of core linalg tests

christiangnrd:
Motivated by attempting to fix OpenCL.jl windows CI (https://github.com/JuliaGPU/OpenCL.jl/pull/405), but the `linalg/core` testset allocates 2x more than any other testsuite, so I think splitting off the two most allocation-heavy testsets into their own testsuites is a good idea regardless.


maleadt (approved):
LGTM. Splitting up these testsets is definitely a good thing.

PR #666, opened 2025-12-07
Contiguous views also for `Base.Slice`

tam724:
Besides `Base.Colon` base also uses `Base.Slice` to indicate complete slices of a dimension (https://github.com/JuliaLang/julia/blob/8fd607a7df9e65ed060c88f32d801c95aa8a573f/base/indices.jl#L382).
For views of GPUArrays, we can treat Slice the same way as Colon and create contiguous arrays (CuArray/.. etc) instead of `SubArray`s.

Closes https://github.com/JuliaGPU/CUDA.jl/issues/2986

### DIFF
--- a/test/testsuite/base.jl
+++ b/test/testsuite/base.jl
@@ -339,11 +339,12 @@ end
         @test_throws BoundsError view(x, :, :, 1:10)
       end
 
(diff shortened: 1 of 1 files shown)


maleadt (approved):
LGTM, thanks!

PR #667, opened 2025-12-12
Add findlast based on findfirst

kshyatt:
Annoying to have only `findfirst` when we do have both `findmin` and `findmax`


kshyatt:
Done! Looks better, imo. I used `Fix1` because it's supported on basically every Julia.

PR #676, opened 2026-01-08
Unbreak mixed type mul on 1.12.4

kshyatt:
Love when LinearAlgebra yet again breaks everything


maleadt (approved):
I hate it. But LGTM, I guess.

PR #679, opened 2026-01-19
Add `diag` argument to `copytri!`

eschnett:
Julia 1.13 added a `diag::Bool` argument to `copytri!`. We need to mimic that so that we can match the signature when overloading for GPUArrays.

### DIFF
--- a/src/host/linalg.jl
+++ b/src/host/linalg.jl
@@ -107,102 +107,48 @@ end
 
 ## copy upper triangle to lower and vice versa
 
-# `v"1.13-"` is the smallest possible 1.13 version number. It comes
-# before all alpha, beta, and rc prereleases. (A plain `v"1.13"` would
-# come after all prereleases.)
-if VERSION >= v"1.13-"
-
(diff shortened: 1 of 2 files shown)


eschnett:
I think runic's indentation is wrong. Look at these lines:
```
        I = @index(Global, Cartesian)
        i, j = Tuple(I)
            if j + diag > i
          @inbounds _A[j,i] = conj(_A[i,j])
        end
```
The `if` statement is spuriously indented too far.

.github/workflows/Test.yml
```diff
@@ -20,7 +20,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        version: ['1.10', '1.11', '1.12', 'nightly']
+        version: ['1.10', '1.11', '1.12', '1.13', 'nightly']
```
maleadt: I think you'll need `1.13-nightly`.

eschnett: Done.

eschnett:
We're going to need to ignore the Buildkite error at the moment. It fails because CUDA.jl isn't working with Julia 1.13. That's a known problem, and a prerequisite for making it work is making GPUArrays.jl work with Julia 1.13...

Issue #677, opened 2026-01-15
(Direct) Constructors for sparse CSC, CSR and BSR arrays


nHackel:
Hello, I think it would be very helpful to add constructors to the generic GPU sparse arrays interface for building sparse matrices directly from index/value GPU vectors, bypassing the COO path used by `sparse`. This would enable packages to precompute the structure and values and construct sparse matrices efficiently across backends.

I was imaging something like:

```julia
# in GPUArrays.jl
export GPUSparseMatrixCSC, ...
function GPUSparseMatrixCSC end
# same for other formats

# in JLArrays.jl
GPUSparseMatrix{Tv, Ti}(rowPtr::JLArray, colVal::JLArray, nzVal::JLArray, dims) = JLSparseMatrixCSC{Tv, Ti}(rowPtr, colVal, nzVal, dims)

# in CUDA.jl
GPUSparseMatrix{Tv, Ti}(rowPtr::CuArray, colVal::CuArray, nzVal::CuArray, dims) = CuSparseMatrixCSC{Tv, Ti}(rowPtr, colVal, nzVal, dims)

# and so on in the other backends
```
One issue is that the arguments might differ between backends or have different requirements on their values(see #648). I think in this case, the formats are so specific though, that the implementations will likely just differ in the types of their vectors.

An example use case can [be seen here](https://github.com/JuliaMath/NFFT.jl/blob/8734f90d894d7875a996776523adb68d710c87fc/ext/NFFTGPUArraysExt/precomputation.jl#L46), where in initial tests the overhead of the COO format negates performance gains over the CPU implementation.

I'm not sure if I fully understand the relationship between `AbstractGPUSparseMatrices` and `GPUSparseDeviceVector` yet and if it would make sense to provide a default implementation for the constructor which defaults to reference to the corresponding default vector.

If this or another version of such a constructor seems sensible, I'd be happy to make PRs

PR #684, opened 2026-01-23
Add direct constructors for sparse arrays

nHackel:
This PR adds direct constructors for the sparse matrix types defined in GPUArrays.jl or rather adds a function for backends to generically define methods for, see also #677.

Changes:
- Introduced a generic function for backends to define constructors for sparse formats.
- Added direct constructors for sparse matrices with a specified format.
- Added a GPUSparseMatrix convenience constructor that defaults to GPUSparseMatrixCOO, which in turn falls back to SparseArrays.sparse for completeness.
- Added tests for CSC and CSR, which can also be extended to the other formats. I was unsure if the constructor tests should also assert behaviour, so far I've only checked for appropriate array types and size.

There currently is no COO and BSR implementation for JLArrays. I could add one as well, at least for the BSR I'd need to take a look at the format first or adapt from maybe CUDA.jl

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,6 +4,7 @@ version = "11.3.4"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
+Atomix = "a9b6321e-bd34-4604-b9c9-b65b8de01458"
 GPUArraysCore = "46192b85-c4d5-4398-a991-12ede77f4527"
 KernelAbstractions = "63c18a36-062a-441e-b654-da1e3ab1ce7c"
 LLVM = "929cbde3-209d-540e-8aea-75f648917ca0"
@@ -24,6 +25,7 @@ JLD2Ext = "JLD2"
 
 [compat]
 Adapt = "4.0"
+Atomix = "1"
 GPUArraysCore = "= 0.2.0"
 JLD2 = "0.4, 0.5, 0.6"
 KernelAbstractions = "0.9.28, 0.10"
--- a/lib/JLArrays/src/JLArrays.jl
+++ b/lib/JLArrays/src/JLArrays.jl
@@ -13,7 +13,7 @@ using GPUArrays
 using Adapt
 using SparseArrays, LinearAlgebra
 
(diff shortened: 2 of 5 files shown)


nHackel:
That sounds like a good middle-ground between the options! I'll try that out as soon as I can

nHackel:
@kshyatt do we want to support a "smart" default , i.e.

```julia
function GPUSparseMatrix(I::AbstractGPUVector, J::AbstractGPUVector, V::AbstractGPUVector, args...; format = default_sparse_format(I))
	if format == :coo
		GPUSparseMatrix(Val(:format), I, J, V)
	elseif format == :csc
		# calculate colptr and use in function call
		GPUSparseMatrix(Val(:format), ...)
	elseif format == :csr
		# calculate rowptr and use in function call
		GPUSparseMatrix(Val(:format), ...)
	# ...
	end
end
```

or a simple default:

```julia
GPUSparseMatrix(I::AbstractGPUVector, J::AbstractGPUVector, V::AbstractGPUVector) = GPUSparseMatrix(Val(:coo), I, J, V)
```

and in either case have:

```julia
GPUSparseMatrix(::Val{:coo}, I, J, V) = GPUSparseMatrixCOO(I, J, V)
GPUSparseMatrix(::Val{:csc}, colPtr, rowVal, nzVal) = GPUSparseMatrixCSC(colPtr, rowVal, nzVal)
GPUSparseMatrix(::Val{:csr}, rowPtr, colVal, nzVal) = GPUSparseMatrixCSR(rowPtr, colVal, nzVal)
GPUSparseMatrix(::Val{:bsr}, ...) = GPUSparseMatrixBSR(...)
```
I'm not sure if the work of implementing the smart default in a GPU compatible manner and the maintenance of it is proportional to the benefit/usage?

nHackel:
I just wanted to comment that I haven't forgotten about this PR, but it's been difficult to find time for open-source work and I'll probably won't find time before JuliaCon/have to put my work-time for open-source things into my conference presentation 😅

kshyatt:
No worries, I think we all know that feeling...

Issue #683, opened 2026-01-21
Test times increased significantly


maleadt:
I updated a local copy of GPUArrays.jl from 97bdfdb to f516fc2, and test times increased from 4 minutes 31 to a whopping 7m32.

Before:

```
[ Info: Running 5 tests in parallel. If this is too many, specify the `--jobs` argument to the tests, or set the JULIA_CPU_THREADS environment variable.
                                                |          | ---------------- CPU ---------------- |
Test                                   (Worker) | Time (s) | GC (s) | GC % | Alloc (MB) | RSS (MB) |
JLArray/alloc cache                         (2) |     1.00 |   0.01 |  0.8 |     192.21 |   557.28 |
JLArray/indexing scalar                     (3) |     5.81 |   0.12 |  2.0 |    1279.38 |   602.16 |
JLArray/indexing find                       (3) |     7.51 |   0.41 |  5.5 |    2602.77 |   765.70 |
JLArray/reductions/reducedim!               (4) |    28.30 |   0.86 |  3.0 |    6716.30 |   662.75 |
JLArray/math/power                          (6) |    29.88 |   1.32 |  4.4 |    7353.23 |   788.39 |
JLArray/reductions/any all count            (4) |     4.22 |   0.10 |  2.4 |    1054.88 |   732.34 |
JLArray/uniformscaling                      (4) |     4.40 |   0.08 |  1.9 |     685.22 |   768.08 |
JLArray/indexing multidimensional           (3) |    24.12 |   0.90 |  3.7 |    4996.91 |   866.39 |
JLArray/linalg/mul!/vector-matrix           (2) |    48.91 |   2.53 |  5.2 |   12632.54 |   814.38 |
JLArray/math/intrinsics                     (2) |     1.55 |   0.00 |  0.0 |     261.91 |   816.22 |
JLArray/reductions/mapreducedim!_large      (4) |    16.79 |   6.39 | 38.0 |   33495.93 |  1238.14 |
JLArray/linalg/NaN_false                    (2) |     9.20 |   0.57 |  6.2 |    2102.50 |   901.80 |
JLArray/statistics                          (2) |    35.87 |   2.51 |  7.0 |    7449.54 |   988.30 |
JLArray/reductions/minimum maximum extrema  (6) |    70.57 |   4.66 |  6.6 |   16305.19 |   992.94 |
[...]

PR #685, opened 2026-01-27
[CI] Use `init_worker_code` to reduce testset initialization overhead

christiangnrd:
See https://github.com/JuliaGPU/GPUArrays.jl/issues/683#issuecomment-3807596624 and https://github.com/JuliaTesting/ParallelTestRunner.jl/pull/85


christiangnrd:
This should wait for https://github.com/JuliaTesting/ParallelTestRunner.jl/pull/88

PR #691, opened 2026-03-03
Use grid-stride loop for `fill!`

christiangnrd:
AMD and Metal do not support total grid size dimensions >=2^32, leading large `fill!`s on AMDGPU to fail to launch (bad), and silently fail on Metal (worse).


vchuravy:
Can we add a check to Metal.jl?

Issue #694, opened 2026-03-27
Indexing (Cartesian or Linear) isn't defined for sparse device arrays


alonsoC1s:
Indexing a CUDA.CUSPARSE with brackets inside a CUDA.jl kernel triggers a `error_if_canonical_getindex` Error. The following reproduces the bug:
```julia
using CUDA, SparseArrays

src_mwe = cu(ones(Float32, 2, 5))
dst_mwe = cu([1, 2, 3, 4, 5])

function mwe_kernel!(dst, src)
    index = threadIdx().x + (blockIdx().x - 1) * blockDim().x
    dst[index] = src[1, index]
    return nothing
end

kernel = @cuda launch=false mwe_kernel!(dst_mwe, sparse(src_mwe, fmt=:csc))
kernel(dst_mwe, sparse(src_mwe, fmt=:csc); threads=length(dst_mwe), blocks=1)

julia> ERROR: a exception was thrown during kernel execution on thread (1, 1, 1) in block (1, 1, 1).
Stacktrace:
 [1] error_if_canonical_getindex at ./abstractarray.jl:1357
 [2] getindex at ./abstractarray.jl:1341
 [3] _getindex at ./abstractarray.jl:1383
 [4] getindex at ./abstractarray.jl:1342
 [5] mwe_kernel! at ./REPL[11]:3
```

The fix is to implement indexing functionality for the given array types in `src/device/sparse.jl`

PR #699, opened 2026-04-01
Implementing scalar indexing for device-side sparse arrays

alonsoC1s:
This PR addresses #694  by implementing scalar indexing for the mentioned array types. Only scalar indexing is implemented,  since doing vectorized indexing would result in dynamic allocations on the device, which is not allowed

Remaining things to be fixed:

- [x] Indexing for `GPUSparseDeviceMatrixCOO` (re-implementing it the way CUDA.jl does it doesn't work for some reason)
- [x] Logical indexing
- [x]  Scalar indexing for BSR formatted arrays
- [x] Including initial indexing tests
- [ ] Add COO, CSR, BSR tests

### DIFF
--- a/src/device/indexing.jl
+++ b/src/device/indexing.jl
@@ -7,7 +7,10 @@ Base.IndexStyle(::Type{GPUSparseDeviceVector}) = Base.IndexLinear()
 # Scalar indexing
 ## Adapted from SparseArrays.AbstractSparseVector
 
-@propagate_inbounds function Base.getindex(v::GPUSparseDeviceVector{Tv,Ti}, i::Integer) where {Tv,Ti}
+@propagate_inbounds function Base.getindex(
+    v::GPUSparseDeviceVector{Tv,Ti},
+    i::Integer,
+) where {Tv,Ti}
     @boundscheck checkbounds(v, i)
     m = nnz(v)
     nzind = nonzeroinds(v)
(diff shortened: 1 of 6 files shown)


kshyatt:
Any chance the rest of the TODOs here could be addressed soon? I'd like to get this in before the next tag :)

alonsoC1s:
@kshyatt of course! I'm working on logical indexing atm. I do need some help to get the tests added (and the checks passing), though

src/device/indexing.jl
```diff
@@ -0,0 +1,152 @@
...
+@propagate_inbounds function Base.getindex(
+    v::GPUSparseDeviceVector{Tv,Ti},
+    i::Integer,
+) where {Tv,Ti}
+    @boundscheck checkbounds(v, i)
+    m = nnz(v)
+    nzind = nonzeroinds(v)
+    nzval = nonzeros(v)
+
+    ii = searchsortedfirst(nzind, convert(Ti, i))
```
kshyatt, line 19: Will this conversion work? You might need to force `i::Ti` above, not sure...

alonsoC1s: I haven't checked with less common number types. This is probably something that would come up naturally during the full testing suite

alonsoC1s:
I finally managed to add initial testing functionality. Notably missing are the tests for any sparse format other than CSC. This is true not only for the proposed changes, but, if I'm not mistaken, formats such as COO are never used in the constructor, linalg, etc. tests

- I changed the indexing implementation to introduce the fixes recently made in CUDA.jl https://github.com/JuliaGPU/CUDA.jl/pull/3101 for relevant types
- From initial testing, the conversion `convert(Ti, i)` poses no problems. Do note that all tests currently only change the eltype (`Int{16, 32, 64}, Float{...}, Complex{...}`), but the index type is always `Int64`

alonsoC1s:
@kshyatt should I open a different PR to introduce the missing sparse types (e.g. COO, BSR) to the whole test suite? Since adding them exclusively to the tests in `sparse.jl` would be more work than adding them across the suite, I feel it would make sense to do it separately

Regarding the tests, since they're build failures on AMD, I'd guess they are unrelated

PR #703, opened 2026-04-09
Store JLArray offset in bytes instead of elements

maleadt:
The element-based offset was lossy when materializing reinterpret on views with non-aligned offsets (e.g., reinterpreting a view of Float32 as Float64). The byte offset would get truncated by integer division when converting to the new element count.

x-ref https://github.com/JuliaGPU/CUDA.jl/issues/2980

### DIFF
--- a/lib/JLArrays/src/JLArrays.jl
+++ b/lib/JLArrays/src/JLArrays.jl
@@ -65,7 +65,7 @@ Adapt.adapt_structure(to::Adaptor, r::Base.RefValue) = JlRefValue(adapt(to, r[])
     Base.sizeof(x::JLDeviceArray) = Base.elsize(x) * length(x)
 
     Base.unsafe_convert(::Type{Ptr{T}}, x::JLDeviceArray{T}) where {T} =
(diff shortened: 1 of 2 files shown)


maleadt:
Other back-ends will suffer from this as well, so will need to be fixed first.

christiangnrd:
The test should probably use `Int16` and `Int32` since some backends don't support `Float64` (oneAPI, Metal) while others don't support `Float16` (current POCL through OpenCL)

maleadt:
Fair point. I've pushed a version doing Int32/Int64, which should always be supported.

This will depend on back-ends getting fixed first, though:
- [x] https://github.com/JuliaGPU/CUDA.jl/pull/3088
- [x] https://github.com/JuliaGPU/Metal.jl/pull/764
- [x] https://github.com/JuliaGPU/oneAPI.jl/pull/570
- [x] https://github.com/JuliaGPU/OpenCL.jl/pull/427
- [x] https://github.com/JuliaGPU/AMDGPU.jl/pull/898

PR #707, opened 2026-04-14
Replace Tausworthe+LCG RNG with stateless Philox4x32-10

maleadt:
Replace the hybrid Tausworthe+LCG generator with a stateless Philox4x32-10 counter-based RNG. Use `@fastmath log` + `sincospi` for Box-Muller randn!, letting each GPU backend provide optimized intrinsics via `@device_override`.

The generic `rand!` uses an immutable ElementRNG struct that implements the minimal AbstractRNG interface (just rand for UInt64 + rng_native_52). Julia's Random stdlib handles all type conversions (Float16, Complex, Bool, integers) automatically. The subcounter state lives in a stack-allocated Ref with a pointer in the struct, avoiding mutable struct heap allocation on GPU.

This was prompted by https://github.com/JuliaGPU/CUDA.jl/issues/3056 / inspired on https://github.com/medyan-dev/PhiloxRNG.jl

Performance is much better than CUDA.jl's NativeRNG, so I'll probably deprecate that in favor of this.

cc @nhz2

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "GPUArrays"
 uuid = "0c68f7d7-f131-5f86-a1c3-88cf8149b2d7"
-version = "11.4.1"
+version = "11.5.0"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
--- a/lib/JLArrays/src/JLArrays.jl
+++ b/lib/JLArrays/src/JLArrays.jl
@@ -539,14 +539,6 @@ end
 
 using Random
 
-const GLOBAL_RNG = Ref{Union{Nothing,GPUArrays.RNG}}(nothing)
-function GPUArrays.default_rng(::Type{<:JLArray})
-    if GLOBAL_RNG[] === nothing
-        GLOBAL_RNG[] = GPUArrays.RNG()
-    end
-    GLOBAL_RNG[]
-end
-
 
 ## GPUArrays interfaces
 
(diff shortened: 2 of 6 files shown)


nhz2:
For `randn` it would be nice to avoid returning exactly zero (though not strictly needed). This is handled by the fixed point math in https://github.com/medyan-dev/PhiloxRNG.jl/blob/main/src/fastsincospi.jl and https://github.com/medyan-dev/PhiloxRNG.jl/blob/main/src/fastlog.jl

The fastsincospi also has a nice exact 8 way symmetry around the unit circle.

These custom functions seem fast on my RTX 3080 with Float32, but maybe in other cases the intrinsics are faster? I haven't looked much into Float64 or other hardware yet.

maleadt:
Yeah I switched back to your versions because `sincospi(::Float32)` uses Float64, which is incompatible with some of our back-ends.

FYI, Claude also came up with a trick to improve Float64 performance:

> **Use bit-pattern trick for `u01(Float64, UInt64)`** — eliminates the
>   `Float64(::UInt64)` conversion + `fma` that hit the 1:64 FP64 wall on
>   consumer GPUs. ~40× speedup for `rand!(F64)`.

Final performance now:

| Benchmark  | Tausworthe (old) | PhiloxRNG (raw) | cuRAND | **this PR** |
|------------|-----------------:|----------------:|-------:|------------:|
| rand! F32  | 138              | 915             | 270    | **912**     |
| rand! F64  | 264<sup>†</sup>  |  21             | —      | **911**     |
| randn! F32 | 170              | 916             | 548    | **912**     |
| randn! F64 |  56              |  12             |  57    |  **97**     |

<sup>†</sup> old Tausworthe F64 uses only 32 bits of entropy per value.

Passes SmallCrush, but I haven't run BigCrush yet.

PR #709, opened 2026-04-16
Fixup/simplify coo_type

kshyatt:


kshyatt:
Checked that this should unblock https://github.com/JuliaGPU/CUDA.jl/pull/3076

kshyatt:
I think the AMD failure is spurious so I'm going to merge

PR #710, opened 2026-04-16
Restore CPU-RNG to GPU-array fallback for rand!/randn!.

maleadt:
Without this, e.g. `randn!(Random.TaskLocalRNG(), ::CuArray)` falls through to the stdlib scalar setindex! path. Regression from #707.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "GPUArrays"
 uuid = "0c68f7d7-f131-5f86-a1c3-88cf8149b2d7"
-version = "11.5.0"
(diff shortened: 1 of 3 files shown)


kshyatt (approved):
Yikes that we gotta copy but I guess that's really the only solution here

maleadt:
It's probably even better for your GPU array-containing struct to provide a `rand!` that uses a GPU-native RNG.

kshyatt:
Yep, that's on the menu for TensorKit as we upgrade its deps to version 6 of everything :D

PR #712, opened 2026-04-19
Use an extra bit of entropy in the Float64 uniform 0-1 distribution.

nhz2:
When looking over https://github.com/JuliaGPU/GPUArrays.jl/pull/707 I think the following is a slight improvement for `Float64` `u01`.
This PR subtracts `prevfloat(1.0)` to avoid returning zero from `u01` while still using 52 bits.

cc @maleadt

### DIFF
--- a/src/host/random.jl
+++ b/src/host/random.jl
@@ -51,10 +51,12 @@ end
 
 @inline function u01(::Type{Float64}, u::UInt64)
     # Bit-pattern construction avoids the expensive Float64(::UInt64) conversion
-    # and fma on consumer GPUs (where FP64 throughput is as low as 1:64). The
-    # low bit of the mantissa is forced set so the result is strictly in (0, 1),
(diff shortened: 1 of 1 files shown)


nhz2:
I think the `Float64(::UInt64)` being expensive was a red herring. The actual issue was the `Float64(2)^(-64)` not being constant propagated.

```julia
@inline function u01(::Type{Float64}, u::UInt64)::Float64
    fma(Float64(u), Float64(2)^Int32(-64), Float64(2)^Int32(-65))
end
```

Seems to be fine performance wise. Ref: https://github.com/medyan-dev/PhiloxRNG.jl/tree/v1.1.1#gpu--nvidia-geforce-rtx-3080-nsvalue-n--100000000

maleadt:
> The actual issue was the `Float64(2)^(-64)` not being constant propagated.

This was fixed in https://github.com/JuliaGPU/CUDA.jl/pull/3098, which I encountered as part of porting your RNG.

PR #715, opened 2026-04-22
Reduce code of testsuite

maleadt:
Trimming useless or redundant test combinations, avoiding needless specialization, etc.


maleadt:
4h50 -> 4h04, that's significant!

PR #717, opened 2026-04-28
Update PhiloxRNG.jl to v1.1.1 and fix counter overflow with UInt64

### LINKED ISSUE #713: Wrap around in `rand_batched_kernel!` can lead to repeating patterns
MWE:

```julia-repl
julia> using GPUArrays, JLArrays, Random

julia> rng = GPUArrays.RNG{JLArray}(42)
GPUArrays.RNG{JLArray}(0x000000000000002a, 0x00000000)

julia> A = JLArray{UInt8}(undef, Int64(2)^34 + 8);

julia> Random.rand!(rng, A);

julia> A[end-7:end] == A[1:8]
true
```

nhz2:
Fixes #713 
This PR uses the following scheme to avoid rng stream collisions.

Philox4x32-10 takes a 128-bit counter `(ctr0, ctr1)` and a 64-bit key. The scheme is:

- **key** = the 64-bit seed (identifies the RNG instance).
- **ctr1** = a 64-bit per-launch counter, incremented host-side after each kernel that consumes the RNG.
- **ctr0** = a 64-bit counter `gid + nthreads * localcounter`, varying inside the kernel: `gid` partitions across threads, `localcounter`  increments per Philox call from that thread. This allows for over 2^32 threads with fewer Philox calls per thread, or over 2^32 Philox calls per thread with fewer threads without having a collision.

`ctr1` is also randomized when seeding from a `RandomDevice` to make collisions less likely. 

CC @maleadt

### DIFF
--- a/src/host/random.jl
+++ b/src/host/random.jl
@@ -76,6 +76,7 @@ end
 #
 # Vendored contents of:
 # https://github.com/medyan-dev/PhiloxRNG.jl/blob/v1.1.1/src/fastsincospi.jl
+# With minor style changes.
 
 # ============================================================
 # Fast sincospi for Box-Muller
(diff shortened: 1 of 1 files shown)


src/host/random.jl
```diff
@@ -69,91 +74,248 @@ end
...
+end
 
-const SP_F32 = (3.1415927f0, -5.167708f0, 2.5497673f0, -0.58907866f0)
-const CP_F32 = (1.0f0, -4.934788f0, 4.057578f0, -1.3061346f0)
+@inline function _cospoly(y::Float64)
+    evalpoly(y * y, _CP64)
+end
 
-@inline function fast_sincospi(::Type{Float32}, u::UInt32)
+@inline function _fast_sincospi(::Type{Float64}, u::Union{UInt32, UInt64})
```
maleadt: Please get rid of Claude'isms like this everywhere. Underscore prefixes are not necessary given `public`/`export` visibility tools.

src/host/random.jl
```diff
@@ -41,22 +41,27 @@ end
...
-    reinterpret(Float64, ((u >> 12) | 0x1) | 0x3ff0000000000000) - 1.0
+Ported from [Random123 uniform.hpp](https://github.com/DEShawResearch/random123/blob/v1.14.0/include/Random123/uniform.hpp#L175).
+"""
+@inline function u01(::Type{F}, u::UInt32)::F where F
+    fma(F(u), F(2)^Int32(-32), F(2)^Int32(-33))
 end
 
+@inline function u01(::Type{F}, u::UInt64)::F where F
+    fma(F(u), F(2)^Int32(-64), F(2)^Int32(-65))
+end
```
maleadt, line 64: Isn't this strictly more expensive than the bit-pattern there was before?

nhz2: I think so, but it has a much larger dynamic range.
See the comment about this in https://github.com/DEShawResearch/random123/blob/9545ff6413f258be2f04c1d319d99aaef7521150/include/Random123/u01fixedpt.h#L43-L52
I also don't see a big impact on benchmarks, https://github.com/medyan-dev/PhiloxRNG.jl#gpu--nvidia-geforce-rtx-3080-nsvalue-n--100000000 RTX 3080 is still almost hitting its memory bandwidth. Does this show up as a problem on other devices?

nhz2: It seems like we can have two variants of this. The bit-pattern version for `rand`, and the `u01` version used internally by `randn`. For `randn` the larger range of `u01` gets translated into better tails.

maleadt: Eh, did some benchmarking myself and couldn't spot a difference. So let's go with the simpler version.

maleadt (approved):

PR #720, opened 2026-05-15
Generalize norm and normalize to AnyGPUArray.

maleadt:
Lets `norm`/`normalize` short-circuit the LinearAlgebra dispatch for SubArrays/wrappers of GPU arrays, which would otherwise fall back to `norm_recursive_check` (Julia 1.12+) or `generic_norm2`, both of which iterate and trigger scalar indexing.

### DIFF
--- a/src/host/linalg.jl
+++ b/src/host/linalg.jl
@@ -757,7 +757,7 @@ end
 
 ## norm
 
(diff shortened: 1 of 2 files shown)


kshyatt:
Does this also cover `norm(v::Broadcasted{<:GPUArrayStyle}, p)` 👀

maleadt:
No, that's not covered by Adapt's `WrappedArray`.

PR #723, opened 2026-05-15
Avoid extra philox calls with ElementRNG

### LINKED ISSUE #721: rand! with RNG and Complex{Int128} isn't random
Only the least significant 64 bits of each Int128 are random. The 64 most significant bits are all zero.

MWE:

```julia-repl
julia> using GPUArrays, JLArrays, Random

julia> rng = GPUArrays.RNG{JLArray}(42)
GPUArrays.RNG{JLArray}(0x000000000000002a, 0x00000000)

julia> A = JLArray{Complex{Int128}}(undef, 1000);

julia> Random.rand!(rng, A);

julia> count(x -> real(x) < 0, A)
0

julia> count(x -> real(x) < 0, rand(Complex{Int128}, 1000))
496

```

### LINKED ISSUE #722: Ambiguity error with rand! and tuple elements.
MWE:

```julia-repl
julia> using GPUArrays, JLArrays, Random

julia> A = JLArray{Tuple{Int32}}(undef, 1000);

julia> rng = GPUArrays.RNG{JLArray}(42)
GPUArrays.RNG{JLArray}(0x000000000000002a, 0x0000000000000000)

julia> Random.rand!(rng, A);
ERROR: MethodError: rand!(::GPUArrays.RNG{JLArray}, ::JLArray{Tuple{Int32}, 1}) is ambiguous.

Candidates:
  rand!(rng::GPUArrays.RNG{AT}, A::AbstractArray{T}) where {AT, T}
    @ GPUArrays ~/.julia/packages/GPUArrays/V2qKA/src/host/random.jl:583
  rand!(rng::AbstractRNG, A::AnyGPUArray)
    @ GPUArrays ~/.julia/packages/GPUArrays/V2qKA/src/host/random.jl:602
To resolve the ambiguity, try making one of the methods more specific, or adding a new method more specific than any of the existing applicable methods.

Stacktrace:
 [1] top-level scope
   @ REPL[20]:1

```

nhz2:
Fixes #721

This PR avoids some extra philox calls when using `ElementRNG`. `ElementRNG` is used as an `AbstractRNG` when sampling any non batched type including types with custom samplers defined in different packages.

This PR also makes `ComplexF16` a batched type.


test/testsuite/random.jl
```diff
@@ -38,6 +38,22 @@ Random.rand(rng::AbstractRNG, ::Random.SamplerType{RGBTriplet}) =
         fill!(A, true)
         rand!(rng, A)
         @test false in Array(A)
+
+        # Int128 is not supported on many backends yet
+        if nameof(AT) == :JLArray
```
maleadt: Ehh, this is ugly. Normally we do this through `eltypes`, which shouldn't list `Int128`.

src/host/random.jl
```diff
@@ -546,12 +542,13 @@ end
...
 # be loaded. Reached via Base's Complex recursion when the element type is
 # e.g. Complex{Float16}.
-@inline Random.randn(rng::ElementRNG, ::Type{Float16}) =
-    first(boxmuller(Float16, rand(rng, UInt32), rand(rng, UInt32)))
-@inline Random.randn(rng::ElementRNG, ::Type{Float32}) =
-    first(boxmuller(Float32, rand(rng, UInt32), rand(rng, UInt32)))
-@inline Random.randn(rng::ElementRNG, ::Type{Float64}) =
-    first(boxmuller(Float64, rand(rng, UInt64), rand(rng, UInt64)))
+for T in (Float16, Float32, Float64, ComplexF32, ComplexF64)
+    @eval @inline function Random.randn(rng::ElementRNG, ::Type{$T})
```
maleadt: How do we reach this for Complex types? ComplexF32 and ComplexF64 are in BatchedRandnTypes. Meanwhile, ComplexF16 isn't, so that's probably the case which should be listed here?

src/host/random.jl
```diff
@@ -416,19 +416,15 @@ end
...
+for T in (UInt64, UInt128, Int128)
+    @eval @inline function Random.rand(rng::ElementRNG, ::Random.SamplerType{$T})
+        sc = unsafe_load(rng.ctr0_ptr) + rng.nthreads
+        unsafe_store!(rng.ctr0_ptr, sc)
+        first(philox_to_vals($T, philox4x32_10(sc, rng.counter, rng.seed)...))
+    end
 end
 
-@inline Random.rand(rng::ElementRNG, ::Random.SamplerType{T}) where T <: Union{Bool,Base.BitInteger} =
+@inline Random.rand(rng::ElementRNG, ::Random.SamplerType{T}) where T <: Union{UInt8, UInt16, UInt32, Int8, Int16, Int32, Int64, Bool} =
```
maleadt: The specific `SamplerType{UInt128}` method should have won over this, so let's keep this short?

src/host/random.jl
nhz2: This can be reached if a type with a custom sampling method calls randn. I can try and also add ComplexF16 to the BatchedRandnTypes.

nhz2:
There are many interesting bugs with using `Float16` and `Int128`, but I don't think they are related to this PR.

nhz2:
Due to https://github.com/JuliaLang/julia/issues/62507 I'm restricting Float16 tests to Julia versions 1.13 or later.

maleadt:
> Due to [JuliaLang/julia#62507](https://github.com/JuliaLang/julia/issues/62507) I'm restricting Float16 tests to Julia versions 1.13 or later.

That's excessive. Float16 had been working fine before; why does it only fail here? If it's specific to a test, you can always add a specific version-gated filter on top of the `eltypes` passed into the testsuite

nhz2:
The Julia 1.12 miscompile only happens on CPUs with avx512_fp16. Github actions will sometimes use a CPU with avx512_fp16 now. If you run the existing tests enough times you can see the failure. I've limited the removal of Float16 tests to only Julia 1.12 and avx512_fp16 CPUs.

maleadt:
The Float16 issue should be solved on 1.12.7, so I removed the workaround. Also rebased the PR.

nhz2:
Thanks. The AMDGPU i128 problem will hopefully be fixed by either https://github.com/llvm/llvm-project/pull/213523 or https://github.com/llvm/llvm-project/pull/214518

maleadt:
> Thanks. The AMDGPU i128 problem will hopefully be fixed by either [llvm/llvm-project#213523](https://github.com/llvm/llvm-project/pull/213523) or [llvm/llvm-project#214518](https://github.com/llvm/llvm-project/pull/214518)

In the interim, let's do https://github.com/JuliaGPU/AMDGPU.jl/pull/1066

PR #731, opened 2026-06-06
test that elementwise ops on mismatched sizes throw DimensionMismatch

JohnCobbler:
Add a guard so mismatched GPU array sizes throw `DimensionMismatch` instead of truncating. Shapes `(10, 1)` and `(3, 1)`, both orders of `+`, plus `.+`.

On CUDA, `+` for a `CuMatrix` is not broadcast. It is cuBLAS `geam`, and `geam!` checks the bounds. `.+` is the `promote_shape` path. CPU backends take `promote_shape` for `+` as well. Same exception either way. See CUDA.jl#2812.

Shared suite, every backend. CUDA.jl#3275 is in, and that broadcasting run was 536/536 with this test included.

### DIFF
--- a/test/testsuite/broadcasting.jl
+++ b/test/testsuite/broadcasting.jl
@@ -132,6 +132,14 @@ function broadcasting(AT, eltypes)
             @test compare(A -> A .* ET(10), AT, rand(ET, 40, 40))
             @test compare((A, B) -> A .* B, AT, rand(ET, 40, 40), rand(ET, 40, 40))
             @test compare((A, B) -> A .* B .+ ET(10), AT, rand(ET, 40, 40), rand(ET, 40, 40))
+
+            @testset "mismatched sizes" begin
+                a = AT(rand(ET, 10, 1))
+                b = AT(rand(ET, 3, 1))
+                @test_throws DimensionMismatch a + b
+                @test_throws DimensionMismatch b + a
+                @test_throws DimensionMismatch a .+ b
+            end
         end
 
         @testset "map! $ET" begin


test/testsuite/broadcasting.jl
```diff
@@ -132,6 +132,14 @@ function broadcasting(AT, eltypes)
             @test compare(A -> A .* ET(10), AT, rand(ET, 40, 40))
             @test compare((A, B) -> A .* B, AT, rand(ET, 40, 40), rand(ET, 40, 40))
             @test compare((A, B) -> A .* B .+ ET(10), AT, rand(ET, 40, 40), rand(ET, 40, 40))
+
+            @testset "mismatched sizes" begin
+                a = AT(rand(ET, 10, 1))
+                b = AT(rand(ET, 3, 1))
+                @test_throws DimensionMismatch a + b
```
aalquwayfili, line 139: These two `+` cases fail on a Quadro RTX 8000 against CUDA.jl main; `a .+ b` below passes. `+` on `CuMatrix` dispatches to cuBLAS `geam`, not Base `promote_shape`, so the symmetric throw comes from the `geam!` bounds check rather than broadcasting.

aalquwayfili (commented):
With JuliaGPU/CUDA.jl#3275 applied the whole broadcasting suite passes (536/536), so this may want to wait for that to land. CI here only runs the CPU backend, so the GPU failure does not show up.

aalquwayfili:
JuliaGPU/CUDA.jl#3275 has merged. Re-ran on a Quadro RTX 8000 against current CUDA.jl main and the broadcasting suite passes, 536/536, including the new testset.

JohnCobbler:
Yep, that's on me. `+` here is geam on CUDA, and the bounds check is what throws; `promote_shape` is the `.+` path. Fixed the description; thanks for confirming.

PR #734, opened 2026-06-10
Keep Symmetric/Hermitian matmul with BLAS eltypes off BLAS.symm!/hemm!.

maleadt:
GPU arrays are DenseArrays, so they match LinearAlgebra's `StridedMatrix{<:BlasFloat}` `generic_matmatmul_wrapper!` methods for the `SymmHemmGeneric` flags, which call `BLAS.symm!`/`hemm!` on device memory. Route them to `generic_matmatmul!` like the existing SyrkHerkGemm override, letting backends (or the generic kernel) handle the wrapper.

Also requires Adapt 4.6.1, which fixes uplo being dropped when adapting Symmetric/Hermitian kernel arguments.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "GPUArrays"
 uuid = "0c68f7d7-f131-5f86-a1c3-88cf8149b2d7"
-version = "11.5.5"
+version = "11.5.6"
 
 [workspace]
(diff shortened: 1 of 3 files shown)


maleadt:
oneAPI.jl failure is real; https://github.com/JuliaGPU/oneAPI.jl/pull/577

PR #738, opened 2026-06-15
Let findmax/min work with Complex and real-valued f

kshyatt:
Right now this errors for complex valued arrays because there's no `typemin`/`typemax` defined for complex numbers, even if `f` argument to `findmax`/`findmin` is real valued.


kshyatt:
Test fails are unrelated so I am going to merge

PR #747, opened 2026-07-02
Fix type usage in opnorm for AbstractGPUSparseMatrixCSC

ChrisRackauckas:
It requires a type not a value. 

Found in https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/28556241206/job/84664223984?pr=3810#step:11:516

### DIFF
--- a/lib/JLArrays/src/JLArrays.jl
+++ b/lib/JLArrays/src/JLArrays.jl
@@ -237,8 +237,8 @@ GPUArrays.dense_array_type(::Type{<:JLSparseMatrixCSC}) = JLArray
 GPUArrays.dense_array_type(sa::JLSparseMatrixCSR) = JLArray
 GPUArrays.dense_array_type(::Type{<:JLSparseMatrixCSR}) = JLArray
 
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
The other possible fix is to add a dispatch to `csr_type(A::cuSPARSE.CuSparseMatrixCSC) = csr_type((typeof(A))`, not sure which is preferred.

kshyatt:
Fails look related now

christiangnrd:
@kshyatt This function is defined according to the definition in the GPUArrays readme in CUDA.jl (takes in the array type), but not JLArrays (takes in the array), but it's used as defined in JLArrays in mapreduce https://github.com/JuliaGPU/GPUArrays.jl/blob/b94388f12d6b83aa537e4890804c7661d9e8706b/src/host/sparse.jl#L1009-L1012

I suggest we fix this in JLArrays and define `csr_type(A<:AbstractGPUSparseMatrix) = csr_type(typeof(A))` (and the `csc_type` equivalent

kshyatt:
Sounds good, should we do this in this PR?

christiangnrd:
> Sounds good, should we do this in this PR?

Yeah I'll push

PR #754, opened 2026-07-28
[JLArrays] allow `mapreducedim!` into wrapper

simeonschaub:
Previously, e.g. `sum!(transpose(v), M)` would error

### DIFF
--- a/lib/JLArrays/Project.toml
+++ b/lib/JLArrays/Project.toml
@@ -1,7 +1,7 @@
 name = "JLArrays"
 uuid = "27aeb0d3-9eb9-45fb-866b-73c2ecf80fcb"
 authors = ["Tim Besard <tim.besard@gmail.com>"]
-version = "0.3.1"
+version = "0.3.2"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
--- a/lib/JLArrays/src/JLArrays.jl
+++ b/lib/JLArrays/src/JLArrays.jl
@@ -555,12 +555,15 @@ else
     end
 end
 
+struct ArrayNoCopy end
+Adapt.adapt_storage(::ArrayNoCopy, x::JLArray) = typed_data(x)
+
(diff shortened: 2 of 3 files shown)


simeonschaub:
Hmm, this is annoying! Looks like all `mapreducedim!` implementations in GPU packages return a `ReshapedArray` with singleton dimensions added, which I'd argue is a bug. Looks like this will need to be fixed in each backend by hand :cry:

christiangnrd:
Can you fix this up into a JLArrays commit and a tests commit and then we won't squash

simeonschaub:
Ok, I can do that! Looks like Metal tests are still failing though, do you have any ideas why?

christiangnrd:
> Ok, I can do that! Looks like Metal tests are still failing though, do you have any ideas why?

It's unrelated to this. You can work around it by using `Base.mapreduce!(identity, +, R, A)` instead of `sum!`

Good catch! https://github.com/JuliaGPU/Metal.jl/issues/907

simeonschaub:
I have pushed a potential fix, which avoids overpromoting in `sum!` and `prod!`. Let's see if this works, then I can clean up the commit history

simeonschaub:
I split it up into two nicely separated commits. Now just needs https://github.com/JuliaGPU/Metal.jl/pull/908 to fix the `sum!` issues.

> You can work around it by using Base.mapreduce!(identity, +, R, A) instead of sum!

I didn't just want to work around it since I think it's good to exercise the `add_sum` path as well for exactly these kinds of issues. Let's see if we can get the Metal fix merged soon, otherwise we can go with the temporary workaround

christiangnrd:
Hey @simeonschaub. I'm planning on releasing 11.5.9 once #755 passes tests, but I think we can get this with the workaround in. Can you edit the JLArrays commit to bump JLArrays to 0.3.2, and the other one to implement the "workaround". And then open a new PR/issue to keep track of the Metal regression?

simeonschaub:
Sure, done!

> And then open a new PR/issue to keep track of the Metal regression?

I've added a reference to https://github.com/JuliaGPU/Metal.jl/issues/907, I think we can just track this there

PR #758, opened 2026-08-09
Align the generic matmul kernels with LinearAlgebra.

maleadt:
#746 widened both factors to the accumulator type to avoid narrow-integer
overflow, but that type isn't a valid conversion target for mixed element
types: `CuMatrix{SVector{2,Float32}} * CuMatrix{Float32}` gives
`Tacc == SVector{2,Float32}`, and `convert(SVector{2,Float32}, ::Float32)`
doesn't exist. Unreachable on the host, a `jl_f_throw_methoderror` call in
device IR. Blocks CUDA.jl CI since 11.5.9 (JuliaGPU/CUDA.jl#3225).

LinearAlgebra doesn't hand-roll this — it accumulates with `muladd`, which
promotes `Number` factors to the accumulator's type (the overflow fix) and
falls back to `x*y + z` for anything else (no conversion attempted). So the
first commit just uses `muladd` and drops `Tacc`.

Testing that turned up two pre-existing bugs in the triangular kernels,
both also a divergence from the stdlib:

- The unit diagonal used `one(Cij)` instead of `oneunit(eltype(A))`, so
  `UpperTriangular(::CuMatrix{Float32}) * ::CuMatrix{SVector{2,Float32}}`
  failed on the missing `one(::SVector)`.
- `C[i,j] += Cij` accumulated into an un-zeroed destination. This one hits
  plain `Float32`: `mul!(JLArray(fill(NaN32,3,3)), UpperTriangular(A), B)`
  returns all NaN. The tests missed it by pre-filling with `zeros`; they
  now use `fill(T(NaN), ...)`.

New `linalg/mul!/mixed-eltype` testsuite entry, using a small
container-like element type so downstream backends need no new dependency.
Full `linalg/*` on `JLArray` and `Array`: 4945 pass, the 16
`Symmetric`/`Hermitian` `uplo=L` failures also fail unmodified. Verified on
a GPU against CUDA.jl.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "GPUArrays"
 uuid = "0c68f7d7-f131-5f86-a1c3-88cf8149b2d7"
-version = "11.5.9"
+version = "11.5.10"
 
 [workspace]
 projects = ["lib/GPUArraysCore", "lib/JLArrays", "test", "docs"]
--- a/src/host/linalg.jl
+++ b/src/host/linalg.jl
@@ -459,10 +459,9 @@ function generic_matmatmul!(C::AbstractArray{R}, A::AbstractArray{T}, B::Abstrac
 
         @inbounds if i <= size(A,1) && j <= size(B,2)
             z2 = zero(A[i, 1]*B[1, j] + A[i, 1]*B[1, j])
-            Tacc = promote_type(R, typeof(z2))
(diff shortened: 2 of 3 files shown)


maleadt:
AMDGPU.jl failure is because https://github.com/JuliaGPU/AMDGPU.jl/pull/1012 hasn't been merged.

PR #760, opened 2026-08-12
Fix multiplication with strided GPU array views

### LINKED ISSUE #756: Matrix products with strided views of JLArrays error since JLArrays 0.3.2 ("Illegal conversion of a JLArray to a Ptr")
Since JLArrays v0.3.2, any `*` / `mul!` involving a strided `SubArray` of a `JLArray` with a BLAS-eligible eltype errors. MWE on Julia 1.12.6, JLArrays v0.3.2:

```julia
using JLArrays, LinearAlgebra
JLArrays.allowscalar(false)

A = JLArray(randn(4, 3))
B = JLArray(randn(4, 5))

view(A, 1:2, :)' * view(B, 1:2, :)
# ERROR: Illegal conversion of a JLArray to a Ptr
# Stacktrace: ... gemm!(...) @ LinearAlgebra.BLAS
```

The same happens for `dense' * view`, `view' * dense`, `view' * vector`, and `mul!` with any strided-view operand. On JLArrays v0.3.1 all of these "worked".

## What changed

`SubArray`s of a `JLArray` with contiguous-ish indices belong to Base's `StridedArray` union (since `AbstractGPUArray <: DenseArray`), so for BLAS eltypes LinearAlgebra routes these products to BLAS `gemm!`/`gemv!`, which calls `unsafe_convert(::Type{Ptr}, ...)`:

- On **v0.3.1**, `Base.unsafe_convert(Ptr{T}, ::JLArray)` silently returned a pointer into the host-backed buffer, so the products ran **CPU BLAS on the emulation array** — numerically correct, but exactly the implicit CPU `ccall` that JLArrays is meant to catch.
- On **v0.3.2**, the new guard (`error("Illegal conversion of a JLArray to a Ptr")`, added alongside the "Pointer access is only available for callers that explicitly want a pointer" change) correctly rejects the conversion — but nothing intercepts the strided-BLAS dispatch before it gets there, so the products now throw.

Non-BLAS eltypes are no better off: `view(Ai, 1:2, :)' * view(Bi, 1:2, :)` with `Int8` falls through to Base's generic matmul and dies with "Scalar indexing is disallowed" under `allowscalar(false)`. So as of 0.3.2, strided views of JLArrays are effectively not multipliable at all.

## Why this matters

JLArrays is the reference backend used downstream to emulate CUDA semantics in CPU CI. On real CUDA, `view(A, 1:2, :)' * view(B, 1:2, :)` works (CUBLAS supports strided device views), so the emulator now rejects code that is fine on the backend it emulates. Downstream example: JuliaGPU-style CI in [cossio/RestrictedBoltzmannMachines.jl](https://github.com/cossio/RestrictedBoltzmannMachines.jl) went red on the JLArrays 0.3.1 → 0.3.2 bump with unchanged code (fixed there by materializing the views: [cossio/RestrictedBoltzmannMachines.jl#206](https://github.com/cossio/RestrictedBoltzmannMachines.jl/pull/206)).

To be clear, the 0.3.2 pointer guard itself looks like the right call — it exposed a real emulation gap that 0.3.1 was silently papering over with host BLAS.

## Suggestion

Would it make sense for GPUArrays/JLArrays to provide `LinearAlgebra.generic_matmatmul!`/`mul!` coverage for strided `SubArray`s (and `Adjoint`/`Transpose` thereof) of `AbstractGPUArray`, routing them to the existing generic GPU matmul kernels, so wrapped views keep working and JLArray matches CuArray behavior? Alternatively, if erroring is the intended semantics for view products on JLArray, a targeted error ("matrix products with views of JLArrays are not supported; materialize with `copy`/`getindex`") would be much easier to act on than the `Ptr` guard firing deep inside BLAS.

maleadt:
Strided views of GPU arrays could incorrectly dispatch to host BLAS, which then failed when attempting to convert GPU storage to a host pointer.

Route these operands through GPUArrays’ generic multiplication kernels instead. Add shared backend coverage for matrix and vector multiplication, symmetric operands, mixed element types, and view outputs. Metal’s native multiplication paths remain unchanged; they already supported strided view inputs to matmul.

Fixes https://github.com/JuliaGPU/GPUArrays.jl/issues/756

### DIFF
--- a/src/host/linalg.jl
+++ b/src/host/linalg.jl
@@ -446,6 +446,67 @@ const AnyStridedGPUArray{T,N} = Union{AbstractGPUArray{T,N},StridedGPUSubArray{T
 const AnyStridedGPUVector{T} = AnyStridedGPUArray{T,1}
 const AnyStridedGPUMatrix{T} = AnyStridedGPUArray{T,2}
 const AnyStridedGPUVecOrMat{T} = Union{AnyStridedGPUVector{T},AnyStridedGPUMatrix{T}}
+const AnyStridedGPUMatrixOperand{T} = Union{
+    AnyStridedGPUMatrix{T},
+    Adjoint{T,<:AnyStridedGPUMatrix{T}},
+    Transpose{T,<:AnyStridedGPUMatrix{T}},
+    Symmetric{T,<:AnyStridedGPUMatrix{T}},
+    Hermitian{T,<:AnyStridedGPUMatrix{T}},
+}
+const AnyStridedGPUVecOrMatOperand{T} = Union{
+    AnyStridedGPUVector{T},
+    AnyStridedGPUMatrixOperand{T},
+}
+
+has_strided_gpu_view(As...) =
+    any(A -> LinearAlgebra._unwrap(A) isa StridedGPUSubArray, As)
+
+# Intercept before LinearAlgebra unwraps operands and dispatches to backend BLAS methods. The
+# signatures also cover view-free products to avoid overlapping methods for each possible view
+# position; those calls are sent back through LinearAlgebra's original implementation.
+@static if VERSION < v"1.11"
+function LinearAlgebra.mul!(C::AnyStridedGPUVector, A::AnyStridedGPUMatrixOperand,
+                            B::AnyStridedGPUVector, a::Number, b::Number)
+    if has_strided_gpu_view(C, A, B)
+        return generic_matmatmul!(C, A, B, a, b)
+    end
+    invoke(LinearAlgebra.mul!,
+           Tuple{AbstractVector,LinearAlgebra.AbstractVecOrMat,AbstractVector,Number,Number},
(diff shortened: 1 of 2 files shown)


christiangnrd:
CUDA tests are failing because it's now attempting to dispatch to `symm!` which CUDA.jl has not definition for Float16.

AMDGPU tests are failing because the `symm!` definitions are only defined for `RocArray`

cossio:
Thanks for fixing this. Could you please tag a new version ?

maleadt:
https://github.com/JuliaRegistries/General/pull/164760

christiangnrd:
I was looking into why the Metal strided array dispatching wasn't showing as covered in codecov and found that 

> Metal’s native multiplication paths remain unchanged; they already supported strided view inputs to matmul.

isn't true and these methods overshadow dispatch to Metal's faster native implementations.

PR #761, opened 2026-08-12
Add a mul!(::AbstractGPUArray, ::Diagonal, ::Diagonal) method

kshyatt:


src/host/linalg.jl
```diff
@@ -294,7 +294,36 @@ function Base.:\(D::Diagonal{<:Any, <:AbstractGPUArray}, B::AbstractGPUVecOrMat)
...
+                            α::Number,
+                            β::Number)
+    dc = view(C, LinearAlgebra.diagind(C))
+    da = A.diag
+    db = B.diag
+    d = length(dc)
+    length(da) == d || throw(DimensionMismatch("right hand side has $(length(da)) rows but output is $d by $d"))
+    length(db) == d || throw(DimensionMismatch("left hand side has $(length(db)) rows but output is $d by $d"))
+    # C may be uninitialized, so a `β` that is zero has to fill rather than scale
+    iszero(β) ? fill!(C, zero(eltype(C))) : rmul!(C, β)
```
christiangnrd, line 323: ~Are there non isbits types `<:Number` that are used with GPUs? If not, `rmul!` should always work right? Or is it a performance thing? If so I would just tweak the comment~

christiangnrd (approved):
One code (comment?) nit.

Also, you could maybe test the beta=0 path but that could also be left to the refactor you `TODO`'d above the "mul! + Diagonal" testset

src/host/linalg.jl
kshyatt: No, the problem is that you can have a situation where you fill `C` with `undef` or `NaN` (like you just allocated it or wrote some crap to it) and if you do `0 * NaN = NaN`.

christiangnrd: Oh right I forgot about `NaN`.

PR #765, opened 2026-08-20
Tighten signatures to avoid method ambiguitity with CUBLAS

kshyatt:
This avoids a method ambiguity error introduced by https://github.com/JuliaGPU/GPUArrays.jl/pull/763, but I'm unsure if it will break things for Metal.


christiangnrd:
You're stuck on an older version of CUDA aren't you?

kshyatt:
Yeah... Enzyme...

christiangnrd:
I think the better fix would be to backport https://github.com/JuliaGPU/CUDA.jl/pull/3240 to CUDA 6.2

kshyatt:
Yeah probably

PR #757, opened 2026-08-09
Define LU factorization for JLArray

ChrisRackauckas-Claude:
Ignore this PR until reviewed by @ChrisRackauckas.

## Summary

JLArrays now defines LinearAlgebra.lu for JLArray matrices by explicitly copying to the host representation before calling host LU. Matching ldiv! methods copy right-hand sides through host memory so the returned factorization can be used by solver code without triggering JLArray to Ptr conversion.

This addresses the regression described in https://github.com/SciML/OrdinaryDiffEq.jl/issues/4179. JLArrays 0.3.2 intentionally rejects implicit CPU pointer conversion, while the generic LinearAlgebra LU path still reaches LAPACK with a JLArray.

## Verification

Before the fix, on the unmodified branch:

    julia --project=test --startup-file=no -e 'using JLArrays, LinearAlgebra; A = jl([2.0 1.0; 1.0 3.0]); lu(A; check = false)'

failed with:

    ERROR: Illegal conversion of a JLArray to a Ptr
    [3] getrf!(A::JLArray{Float64, 2}, ipiv::JLArray{Int64, 1}; check::Bool)
        @ LinearAlgebra.LAPACK .../LinearAlgebra/src/lapack.jl:587

With this patch:

    julia --project=test --startup-file=no test/jlarrays_lu.jl

passed:

    Test Summary: | Pass  Total  Time
    JLArray LU    |    3      3  1.3s

The OrdinaryDiffEq reproduction also passed for both TRBDF2 and Rodas5P:

    (SciMLBase.ReturnCode.Success, SciMLBase.ReturnCode.Success, [0.04713496106370036, 0.04713496106370036, 0.04713496106370036])

The full GPUArrays suite passed:

    Test Summary: |  Pass  Total      Time
      Overall     | 21022  21022  16m23.0s
       SUCCESS
        Testing GPUArrays tests passed

Also ran typos on the changed files and git diff --check. GPU-specific backends and downstream packages were not tested.

PR #769, opened 2026-08-24
[JLArrays] Revert a couple changes to unbreak downstream CI

christiangnrd:
The KA 0.10 interface has changed so it's no longer true. Also revert the pointer change to be reapplied in a breaking release of JLArrays

Unlike the real GPU backends, JLArrays support for KernelInterface will depend on KA 0.10 since it'll rely on the new POCLBackend

Should wait for https://github.com/JuliaRegistries/General/pull/165667 before merging/registering

Close #757

### DIFF
--- a/lib/JLArrays/Project.toml
+++ b/lib/JLArrays/Project.toml
@@ -1,7 +1,7 @@
 name = "JLArrays"
 uuid = "27aeb0d3-9eb9-45fb-866b-73c2ecf80fcb"
-authors = ["Tim Besard <tim.besard@gmail.com>"]
 version = "0.3.2"
+authors = ["Tim Besard <tim.besard@gmail.com>"]
 
 [deps]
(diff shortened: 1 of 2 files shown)


maleadt:
> Also revert the pointer change to be reapplied in a breaking release of JLArrays

Why is this breaking? If anything it should indicate things we're missing to implement, it's not breaking wrt. the user API.

christiangnrd:
It's breaking in a "this bug is now expected and worked around so fixing it would break user code" kind of way and I've seen 2 instances of test suites breaking. (https://github.com/JuliaGPU/GPUArrays.jl/pull/757 and another I can't remember)

The proper fix would be to add new fallback kernels in GPUArrays but I don't have time to do that at the moment so I think reverting for JLArrays 0.3 and then setting the expectation that in JLArrays 0.4, things will break if they're not supported by GPUArrays.

PR #784, opened 2026-09-21
Identify shared GPU allocations in Base.dataids

### LINKED ISSUE #716: Aliasing with `JLArray`s and broadcasting
Bumped into a fun subtle issue where broadcasting with `JLArray`s and views behaves differently from their `Array` counterparts.

The specific case is this:

```julia
julia> using JLArrays

julia> A = -jl(ones(3, 3));

julia> A .*= sign.(view(A, 1:4:9)) # incorrect result!
3×3 JLArray{Float64, 2}:
 1.0  -1.0  -1.0
 1.0   1.0  -1.0
 1.0   1.0   1.0

julia> B = -ones(3, 3);

julia> B .*= sign.(view(B, 1:4:9)) # correct result!
3×3 Matrix{Float64}:
 1.0  1.0  1.0
 1.0  1.0  1.0
 1.0  1.0  1.0
```

From what I understand, the issue here is that broadcasting has some machinery to detect aliasing in its in/outputs, which ensures that the second case (`B`) works (I didn't delve deeply into this, but I think an independent copy is made of the view which then ensures writing into `B` does not alter reading from the view).
This machinery doesn't seem to propagate fully through the view of the `JLArray` though.

This came up in https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/pull/219.
To be honest, I'm not sure if this is really a case of "bug" or just more of a wrong usage from my part, as I actually hadn't intended for the view and the output to share data.
I was more surprised that this wasn't noticed in our tests for `Array`s because there this is caught, while here it isn't.
I mostly just wanted to report this in case this wasn't known.

YidaWeng:
Fixes #716.

In-place broadcast of a GPU array with a strided view of itself, such as `A .*= sign.(view(A, 1:4:9))`, returned the wrong values. `Base.unalias` treated the view as independent memory, so the kernel read entries that the same broadcast had already overwritten.

`view` on an `AbstractGPUArray` reshapes the parent first. The view's parent is therefore a new wrapper around the same `DataRef`, and the default `dataids` is `objectid` of that wrapper. `copy(::DataRef)` shares the `RefCounted` record, so `dataids` now uses `objectid` of that record for both `DataRef` and `AbstractGPUArray`.

Backends that already define a more specific `dataids` (CUDA and oneAPI use a pointer range, and override `mightalias`) keep those methods. This change covers backends whose storage is a `DataRef` and that do not define their own `dataids`, including `JLArray` and current `MtlArray`.

## Tests

- Regression in `test/testsuite/broadcasting.jl`: `mightalias` for a strided view and a reshape, plus the in-place broadcast from the issue, compared with `Array`.
- Full `Pkg.test()` on Julia 1.13.0, macOS (Apple Silicon): 23838 passed, covering `JLArray` and `Array`.

No CUDA, AMDGPU, oneAPI, Metal, or OpenCL run locally. Metal should pick this up because `MtlArray` stores a `DataRef` and does not define `dataids`. CUDA and oneAPI keep their existing pointer-range checks.


YidaWeng:
Buildkite [build 1937](https://buildkite.com/julialang/gpuarrays-dot-jl/builds/1937/list) failed on all Metal and oneAPI Julia versions (8 jobs). CUDA and AMDGPU passed.

The `aliased view` regression used `ones(3, 3)` (`Float64`). Metal reports `Metal does not support Float64 values`; oneAPI reports `Float64 is not supported on this device`. The dataids change itself was not the failure.

Pushed a follow-up that picks `Float32` from the backend's `eltypes` list.

